PR #19, opened 2023-08-07
Upgrade rustls

gaojiaqi7:
The latest `async-rustls` relies on rustls `0.21` codebase. If we want to utilize this crate, we need to update the `rustls` submiodule.

However, due to a `no_std` issue that has just been fixed in `rustls-webpki`, we have to use the latest `rustls` main branch. We can switch to a release tag once it has a new release.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -30,15 +30,15 @@ dependencies = [
 
 [[package]]
 name = "anstyle"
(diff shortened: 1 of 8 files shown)


jyao1 (approved):

jyao1:
Please fix CI failure.

Issue #22, opened 2023-08-09
migtd default policy for SVN does not work


jyao1:
https://github.com/intel/MigTD/blob/main/config/policy.json

Currently, below is using greater-or-equal.
```
            "TEE_TCB_SVN.SEAM":{
                "operation":"greater-or-equal",
                "reference":"self"
            },
```

The meaning is: the data in **destination** is greater-or-equal to the data in **source**.
But the current flow is the data in **peer** is greater-or-equal to the data in **local**. (https://github.com/intel/MigTD/blob/main/src/policy/src/config.rs#L132)

gaojiaqi7:
if the `reference` is a numerical value and `operation` is `greater-or-equal`, it means `peer >= value` or `dst >= value`?

PR #27, opened 2023-08-11
policy: compare dst and src when op is greater_or_equal

gaojiaqi7:
When the operation of a property in policy is greater or equal and the reference value if `self`, we should compare between dst and src, otherwise the only qualified value will be equal to `self`.

Since MigTD-dst always start a TLS server and MigTD-src always start a TLS client, we can distinguish between migtd-s and d in this way when verifying the policy.

Fix: https://github.com/intel/MigTD/issues/22

### DIFF
--- a/src/migtd/src/mig_policy.rs
+++ b/src/migtd/src/mig_policy.rs
@@ -8,7 +8,11 @@ use tdx_tdcall::tdreport;
 
 use crate::{config::get_policy, event_log::get_event_log};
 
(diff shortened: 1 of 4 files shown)


jyao1:
@liuw1 , do we have integration test need to fix or add?

liuw1:
> @liuw1 , do we have integration test need to fix or add?

Yes, I submitted #28 to track this task.

jyao1 (approved):

Issue #24, opened 2023-08-09
vsock fuzzing is broken


gaojiaqi7:
It is closed for now, we need to reopen it once we fix the fuzzing case
https://github.com/intel/MigTD/blob/0173118898a5d5b1dd8687400b4f1951fca4cf80/src/devices/vsock/fuzz/Cargo.toml#L33-L37

PR #34, opened 2023-08-29
devices/vsock: fix vsock fuzz test

gaojiaqi7:
Fix: https://github.com/intel/MigTD/issues/24


src/devices/vsock/src/transport/event.rs
```diff
@@ -34,3 +35,20 @@ pub fn wait_for_event(event_flag: &AtomicBool, timer: &dyn VsockTimeout) -> bool
 
     true
 }
+
+#[cfg(feature = "fuzz")]
+pub fn wait_for_event(event_flag: &AtomicBool, timer: &dyn VsockTimeout) -> bool {
```
jyao1, line 40: Can we merge those 2 functions?

gaojiaqi7: Done

Issue #54, opened 2023-09-27
virtio-serial: re-initiate premigration fails with the previous error status


wei-w-wang:
When pre-migration fails with something missed in the setup, after fixing the setup and re-initiate premigration, it still always fails with the previous error status. The WA currently is to destroy the src and dst MigTDs and re-start them again.

Reproduce steps:
1. boot src and dst MigTDs
2. boot src user TD (with the pre-binding setup)
3. run the premig.sh script (commands to do binding and initiate pre-migration)
4. Host reports pre-migration failed, state=6
The above is expected, as we forgot to launch the dst user TD in setp 2

The issue is: after we fix setp 2 with launching both the src and dst user TD, and re-do step 3, pre-migration still reports an error. This works fine with the legacy virtio-vsock based pre-migration. Probably some status maintained in MigTD needs to be reset after re-binding happens.

gaojiaqi7:
The root cause is, at the first time, `MigTD-src` is waked up to try to establish TLS connnection but `MigTD-dst` is not. The TLS packet sent from `dst` will still be sent to `src` and be consumed in the TLS handshake of the `next` request.

The solution will be: `MigTD` open the serial port after receiving a pre-migration request and close the port after the request is processed.

PR #59, opened 2023-10-16
Resolve the redundant tls packet issue

gaojiaqi7:
Fix https://github.com/intel/MigTD/issues/54


jyao1 (approved):

jyao1:
@wei-w-wang, would you please review that?

wei-w-wang:
> @wei-w-wang, would you please review that?

This seems to be the simplest way we have. We can use it for now.

Issue #10, opened 2023-06-21
need calculate memory usage for multiple connection 


gaojiaqi7:

gaojiaqi7:
| | Heap | Stack | DMA |
| --- | --- | --- | --- |
| No request | 533,688 B | 12,360 B | 98,304 B |
| 1 requests | 690,635 B | 47,064 B | 159,744 B |
| 10 requests | 2,103,158 B | 359,400 B | 1,376,256 B |
| 16 request | 3,044,840 B | 567,624 B | 2,187,264 B |

gaojiaqi7:
The currenct metadata configuration (32M perm memory) can support up to ~50 requests concurrently per calculation.

jyao1:
Current code is using `alloc`.

Please evaluate the impact - either create a method to calculate the memory size or remove the alloc to make the calculation simpler.

gaojiaqi7:
Close with https://github.com/intel/MigTD/issues/458

PR #55, opened 2023-10-09
Introduce async crates

gaojiaqi7:

PR #70, opened 2023-10-31
migtd: enlarge the runtime layout based on the metadata

gaojiaqi7:
Enlarge the stack, heap and DMA size to fully use the usable memory.

Closes: https://github.com/intel/MigTD/issues/10


jyao1:
why enlarge?
do we still need this?

gaojiaqi7:
The runtime memory is not used up, so the stack, heap and DMA regions can be enlarged to support more sessions in async scenario https://github.com/intel/MigTD/pull/55

jyao1:
please resolve conflict.

Issue #71, opened 2023-10-31
change name DMA to shared memory


jyao1:

gaojiaqi7:
Waiting for PR https://github.com/confidential-containers/td-shim/pull/616 being merged

OuyangHang33:
Here, I'm checking this issue.

liuw1:
There are many dma words in virtio related code, do we need to change name?

jyao1:
yes remove DMA keyword

gaojiaqi7:
Closed byhttps://github.com/intel/MigTD/pull/92

PR #76, opened 2023-11-08
Update TD-shim patch and change shared memory

OuyangHang33:
Update TD-shim patch
Td-payload: change Dma to SharedMemory
Fix: #71


gaojiaqi7:
Closes this PR which duplicates with https://github.com/intel/MigTD/pull/92

Issue #86, opened 2023-11-20
need a way to pre-calculate TEE_INFO_HASH of MigTD


jyao1:
As such, the user can pre-bind the MigTD SERVTD_INFO_HASH

If CFV is needed, then we need a way to allow the tool provision CFV measurement to RTMR.

PR #99, opened 2023-12-12
tools: introduce `servtd-hash`

gaojiaqi7:
`sertd-hash` is used to calculate `TEE_INFO_HASH` of migtd using migtd image and configuration manifest.

`td-shim/td-shim-tools` is used to calculate mrtd and rtmr[0-1] and extended to calculate the rtmr2 which is used by migtd.

Closes https://github.com/intel/MigTD/issues/86

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -14,6 +14,15 @@ dependencies = [
 ]
 
 [[package]]
+name = "aho-corasick"
+version = "1.1.2"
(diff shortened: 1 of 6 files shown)


jyao1:
why we cannot use the tool in td-shim?

jyao1:
Suggest to change `servtd-hash` to `migtd-hash`

jyao1 (approved):

PR #155, opened 2024-01-29
attestation: update to latest linux-sgx branch

gaojiaqi7:
Closes https://github.com/intel/MigTD/issues/154


jyao1:
What is the difference between original code to gen .a lib and the new code?

jyao1:
do you need to update gitsubmodule?

gaojiaqi7:
It has been updated in the commit

gaojiaqi7:
> What is the difference between original code to gen .a lib and the new code?

There is no source code difference between the existing static library and the new branch.

jyao1 (approved):

Issue #103, opened 2023-12-13
set MIG_PROTOCOL for cross platform migration


jyao1:
To support cross platform migration, MigTD shall read the `MIN/MAX_EXPORT/IMPORT_VERSION` using `TDG.SYS.RD`, and set `MIG_VERSION` using `TDG.SERVD.WR`.

PR #158, opened 2024-02-07
migtd: set `MIG_VERSION` before programming MSK

gaojiaqi7:
MigTD-s and MigTD-d exchange their MSK together with the version information and then they will evaluate the versions.

If there are one or more versions that meet both requirements, they will set the maximum one, otherwise the pre-migration flow will abort.

Closes https://github.com/intel/MigTD/issues/103


jyao1:
Please describe what test has been done for this patch

gaojiaqi7:
Integration test passed. Unit test is added to check the calculated `MIG_VERSION`

Issue #87, opened 2023-11-20
why use vec to hold the GHCI communciation data


jyao1:
https://github.com/intel/MigTD/blob/main/src/migtd/src/migration/session.rs#L399 

Vector means the size is unpredictable.
I think it should use the static location.

gaojiaqi7:
@jyao1 I think the size of vector used here is predictable because the size is equal to the allocated shared memory which is fixed. The return results from VMM do not affect the memory allocation.

PR #159, opened 2024-02-22
migtd: copy GHCI data to stack memory instead of heap

gaojiaqi7:
Closes https://github.com/intel/MigTD/issues/87


src/migtd/src/migration/session.rs
```diff
@@ -89,7 +90,8 @@ impl MigrationSession {
         #[cfg(not(feature = "vmcall-interrupt"))]
         tdx::tdvmcall_service(cmd_mem.as_bytes(), rsp_mem.as_mut_bytes(), 0, 0)?;
 
-        let private_mem = Self::copy_from_shared_memory(rsp_mem.as_bytes());
+        let mut private_mem = [0u8; GHCI_PAGE_SIZE];
```
jyao1, line 93: you can use a global data section instead of stack.

PR #181, opened 2024-03-06
Fix debug migtd image build

gaojiaqi7:
- Adjust the image layout to make enough space for `Payload`
  - Add debug migtd build job in CI

### DIFF
--- a/src/migtd/src/config.rs
+++ b/src/migtd/src/config.rs
@@ -3,10 +3,11 @@
 // SPDX-License-Identifier: BSD-2-Clause-Patent
 
 use r_efi::efi::Guid;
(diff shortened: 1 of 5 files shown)


src/migtd/src/config.rs
```diff
@@ -6,7 +6,7 @@ use r_efi::efi::Guid;
 use td_uefi_pi::{fv, pi};
 
 pub const CONFIG_VOLUME_BASE: usize = 0xFF000000;
-pub const CONFIG_VOLUME_SIZE: usize = 0x40000;
+pub const CONFIG_VOLUME_SIZE: usize = 0x20000;
```
jyao1, line 9: why we need change this? Everything should be in image_layout.json.

jyao1 (approved):

PR #184, opened 2024-03-07
Add log level option at compile time

gaojiaqi7:
Closes https://github.com/intel/MigTD/issues/182

### DIFF
--- a/src/migtd/src/bin/migtd/main.rs
+++ b/src/migtd/src/bin/migtd/main.rs
@@ -7,9 +7,9 @@
 
 extern crate alloc;
 
+use log::info;
 use migtd::migration::{session::MigrationSession, MigrationResult};
 use migtd::{config, event_log, migration};
-use td_payload::println;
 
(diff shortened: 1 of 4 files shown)


src/migtd/src/bin/migtd/main.rs
```diff
@@ -103,7 +103,7 @@ fn handle_pre_mig() {
                 let coverage_len = minicov::get_coverage_data_size();
                 assert!(coverage_len < MAX_COVERAGE_DATA_PAGE_COUNT * td_paging::PAGE_SIZE);
                 minicov::capture_coverage_to_buffer(&mut buffer[0..coverage_len]);
-                println!(
+                td_payload::println!(
```
jyao1, line 106: why use td_payload::println? but not info! or error! ?

gaojiaqi7: Since it is test code for benchmark, maybe it will be better that it is not controlled by log level.

jyao1 (approved):

PR #189, opened 2024-03-12
Close warnings reported by Clippy

gaojiaqi7:
Warning types:
 - the Err-variant returned from this function is very large
 - unused import
 - use of a fallible conversion when an infallible one could be used
 - the borrowed expression implements the required traits
 - all variants have the same prefix
 - comparing with null is better expressed by the .is_null() method
 - casting to the same type is unnecessary
 - this expression creates a reference which is immediately dereferenced by the compiler
 - this impl can be derived

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -760,6 +760,7 @@ dependencies = [
  "cc-measurement",
  "crypto",
  "lexical-core",
+ "log",
  "serde",
  "serde_json",
  "td-shim",
--- a/src/attestation/build.rs
+++ b/src/attestation/build.rs
@@ -30,13 +30,13 @@ fn main() {
 
     // make servtd_attest_preparation
     Command::new("make")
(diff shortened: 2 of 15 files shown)


liuw1 (approved):

src/migtd/src/mig_policy.rs
```diff
@@ -12,17 +13,17 @@ pub fn authenticate_policy(
     verified_report_local: &[u8],
     verified_report_peer: &[u8],
     event_log_peer: &[u8],
-) -> Result<(), PolicyError> {
+) -> Result<(), Box<PolicyError>> {
```
jyao1, line 16: why use Box?

gaojiaqi7: Removed this by logging error information inside policy crate without returning out to caller

jyao1 (approved):

PR #202, opened 2024-04-09
migtd: mask the `xfam` and `attribute`

gaojiaqi7:
MigTD may have different features on different platforms. Mask the attribute and xfam bits that is not required for migtd.

### DIFF
--- a/src/policy/src/verify.rs
+++ b/src/policy/src/verify.rs
@@ -24,6 +24,40 @@ const EV_EVENT_TAG: u32 = 0x00000006;
 const TAGGED_EVENT_ID_POLICY: u32 = 0x1;
 const TAGGED_EVENT_ID_ROOT_CA: u32 = 0x2;
 
+// Attributes Mask:
+// Bits     Feature             Masked or Not   Notes
+// 3:0      TD Under Debug      Not masked      Always 0
+// 15:4     TD Under Profiling  Not masked      Always 0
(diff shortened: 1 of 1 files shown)


src/migtd/src/ratls/server_client.rs
```diff
@@ -123,7 +123,7 @@ fn verify_peer_cert(
     cert: &[u8],
     quote_local: &[u8],
 ) -> core::result::Result<(), CryptoError> {
-    let verified_report_local = attestation::verify_quote(quote_local)
+    let mut verified_report_local = attestation::verify_quote(quote_local)
```
jyao1, line 126: please don't change attribute here.

We can have a local mut version in authenticate_policy().

src/migtd/src/ratls/server_client.rs
```diff
@@ -137,14 +137,14 @@ fn verify_peer_cert(
         parse_extensions(extensions).ok_or(CryptoError::ParseCertificate)?;
 
     #[cfg(not(feature = "test_disable_ra_and_accept_all"))]
-    if let Ok(verified_report_peer) = attestation::verify_quote(quote_report) {
+    if let Ok(mut verified_report_peer) = attestation::verify_quote(quote_report) {
```
jyao1, line 140: Ditto.

jyao1 (approved):

PR #207, opened 2024-04-16
virtio: fix index check

### LINKED ISSUE #205: The index check should be `index >= 6` in set_region
The index check should be `index >= 6` because the array has 6 elements (0-5)
```
impl CheckRegionOverlap {
    fn set_region(&mut self, index: usize, base: u64, length: u64) -> Result<()> {
        if index > 6 {
            return Err(VirtioError::InvalidParameter);
        }

        self.flag[index] = true;
        self.interval[index][0] = base;
        self.interval[index][1] = base + length;

        Ok(())
    }
}
```

liuw1:
Fix #205

### DIFF
--- a/src/devices/virtio/src/virtio_pci.rs
+++ b/src/devices/virtio/src/virtio_pci.rs
@@ -81,7 +81,7 @@ struct CheckRegionOverlap {
 
 impl CheckRegionOverlap {
     fn set_region(&mut self, index: usize, base: u64, length: u64) -> Result<()> {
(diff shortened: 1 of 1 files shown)


gaojiaqi7:
The `VirtioPciCapabilityType` has maximum value `5`, so the index boundary should not be changed

liuw1:
> VirtioPciCapabilityType

OK，so please add a PR to replace hardcode 6.

PR #273, opened 2024-07-23
Update pre_production policy

liuw1:


liuw1:
WW30.2, I used [migtd-policy-generator](https://github.com/intel/MigTD/blob/main/tools/migtd-policy-generator/readme.md) tool to generate latest policy file. A new fmspc "10A06E050000" has been updated in SBX server.

jyao1 (approved):

PR #276, opened 2024-07-23
Enable trivy scan in CI

liuw1:


.github/workflows/trivy.yml
```diff
@@ -0,0 +1,47 @@
...
+        uses: actions/checkout@v4
+        with:
+          submodules: recursive
+
+      - name: Run Trivy vulnerability scanner
+        uses: aquasecurity/trivy-action@7b7aa264d83dc58691451798b4d117d53d21edfe
+        with:
+          scan-type: 'fs'
+          scanners: 'vuln'
+          ignore-unfixed: true
```
xiaoyuxlu, line 39: why ignore unfixed?

.github/workflows/trivy.yml
```diff
@@ -0,0 +1,47 @@
...
+    name: Build
+    runs-on: "ubuntu-20.04"
+    steps:
+      - name: Checkout code
+        uses: actions/checkout@v4
+        with:
+          submodules: recursive
+
+      - name: Run Trivy vulnerability scanner
+        uses: aquasecurity/trivy-action@7b7aa264d83dc58691451798b4d117d53d21edfe
```
xiaoyuxlu, line 35: why use this commit( 2 years ago)? why not use the latest one (0.24.0 2weeks ago)?

.github/workflows/trivy.yml
liuw1: Removed

.github/workflows/trivy.yml
liuw1: Good catch, updated.

jyao1 (approved):

PR #282, opened 2024-08-01
Fix MSI-X table offset by masking off the BIR bits

BenDHillier:
The MSI-X table offset is supposed to have the BIR bits masked off. PCI Spec V3.0 6.8.2.4: "The lower 3 Table BIR bits are masked off (set to zero) by software to form a 32-bit QWORD-aligned offset".


gaojiaqi7 (approved):
LGTM. Thanks @BenDHillier for the fix

PR #320, opened 2024-11-15
Add multiple migration support for `vmcall-vsock`

gaojiaqi7:


src/migtd/src/driver/timer.rs
```diff
@@ -32,7 +27,7 @@ pub fn schedule_timeout(timeout: u64) -> Option<u64> {
     reset_timer();
     let cpuid = unsafe { core::arch::x86_64::__cpuid_count(0x15, 0) };
     let tsc_frequency = cpuid.ecx * (cpuid.ebx / cpuid.eax);
-    let timeout = timeout / 1000 * tsc_frequency as u64;
+    let timeout = tsc_frequency as u64 * timeout / 1000;
```
jyao1, line 30: why change this?

src/migtd/src/bin/migtd/main.rs
```diff
@@ -135,6 +137,50 @@ fn handle_pre_mig() {
     }
 }
 
+#[cfg(feature = "async")]
+fn handle_pre_mig() {
```
jyao1, line 141: why we need to handle `sync` and `async` differently? Why we need a feature?

src/migtd/src/migration/session.rs
```diff
@@ -155,7 +162,7 @@ impl MigrationSession {
         Ok(())
     }
 
-    pub fn wait_for_request(&mut self) -> Result<()> {
+    pub fn wait_for_request(&mut self, block: bool) -> Result<Option<u64>> {
```
jyao1, line 165: why we need to support `block` here? Can we use false for original use case?

src/migtd/src/bin/migtd/main.rs
gaojiaqi7: The original `wait_for_request` on Linux blocks the program from reading/writing from virtio vsock/serial until next migration request comes. I suggest to file a new issue to track the async compatiblity on block version `wait_for_request`.

src/migtd/src/driver/timer.rs
gaojiaqi7: For 1, the purpose of this line of code is to calculate the increase in the Time Stamp Counter (TSC) value over a specified delay period, I think rewrite as `tsc_frequency as u64 / 1000 * timeout as u64` is a better way to address the potential overflow.

For 2/3/4, the code/comments has been updated.

src/migtd/src/migration/session.rs
gaojiaqi7: Removed. The async and sync code blocks are merged.

PR #337, opened 2025-01-24
Check exit status of build script commands.

davidskidmore:
The result from running a command only indicates whether it was able to spawn and does not check whether it finished successfully. This change is to check whether the exit status was successful so `make` failures are not silently ignored.

Signed-off-by: David Skidmore <davidskidmore@google.com>

### DIFF
--- a/src/attestation/build.rs
+++ b/src/attestation/build.rs
@@ -29,16 +29,21 @@ fn main() {
         .to_string();
 
     // make servtd_attest_preparation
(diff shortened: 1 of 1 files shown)


jyao1:
@davidskidmore , would you please fix CI issue?

davidskidmore:
I fixed the formatting issue, but there is an integration test failure now.

jyao1 (approved):

PR #349, opened 2025-03-13
vmcall-vsock: always return false in `can_recv`

gaojiaqi7:
Unlike virtio-vsock, the can_recv method of vmcall-vsock should return false because each recv operation depends on a vmcall.


src/devices/vsock/src/transport/vmcall.rs
```diff
@@ -306,7 +306,7 @@ impl VsockTransport for VmcallVsock {
 
     /// Whether can receive packet.
     fn can_recv(&self) -> bool {
-        true
+        false
```
jyao1, line 309: please add comment here to explain why.

jyao1 (approved):

PR #379, opened 2025-04-25
migtd: ghci 1.5 spec implementation

mjjagasi:
migtd: ghci 1.5 spec implementation

Signed-off-by: Mukesh Jagasia <mukesh.j.jagasia@intel.com>

### DIFF
--- a/src/devices/vmcall_raw/src/lib.rs
+++ b/src/devices/vmcall_raw/src/lib.rs
@@ -1,0 +1,129 @@
+// Copyright (c) 2025 Intel Corporation
+//
+// SPDX-License-Identifier: BSD-2-Clause-Patent
+
+#![cfg_attr(not(test), no_std)]
+
+extern crate alloc;
+use core::fmt::{self, Display};
+use rust_std_stub::{error, io};
+use tdx_tdcall::TdVmcallError;
+
+pub mod stream;
+pub mod transport;
+
+const PAGE_SIZE: usize = 0x1000;
+
+#[derive(Debug)]
+pub enum VmcallRawError {
+    /// Initialization error
+    Initialization,
+    /// Device not available
+    DeviceNotAvailable,
+    /// Cannot allocate unused port
+    NoAvailablePort,
+    /// Port has been already used
+    AddressAlreadyUsed,
+    /// Tranport: device io error
+    TransportError,
+    /// Packet buffer is too short.
+    Truncated,
+    /// Packet header can not be recognized.
+    Malformed,
+    /// An operation is not permitted in the current state.
+    Illegal,
+    /// There is no listen socket on remote
+    REFUSED,
+    /// There is no data has been sent or received
+    NotReady,
+    /// Tdvmcall error
+    TdVmcallErr,
+    /// Interrupt error
+    Interrupt,
+}
+
+impl Display for VmcallRawError {
+    fn fmt(&self, f: &mut fmt::Formatter) -> fmt::Result {
+        match self {
+            VmcallRawError::Initialization => write!(f, "Initialization"),
+            VmcallRawError::DeviceNotAvailable => write!(f, "DeviceNotAvailable"),
+            VmcallRawError::TransportError => write!(f, "TransportError"),
+            VmcallRawError::Truncated => write!(f, "Truncated"),
+            VmcallRawError::Malformed => write!(f, "Malformed"),
+            VmcallRawError::Illegal => write!(f, "Illegal"),
+            VmcallRawError::REFUSED => write!(f, "REFUSED"),
+            VmcallRawError::NoAvailablePort => write!(f, "NoAvailablePort"),
+            VmcallRawError::AddressAlreadyUsed => write!(f, "AddressAlreadyUsed"),
+            VmcallRawError::NotReady => write!(f, "NotReady"),
+            VmcallRawError::TdVmcallErr => write!(f, "TdVmcallErr"),
+            VmcallRawError::Interrupt => write!(f, "Interrupt"),
+        }
+    }
+}
+
+impl From<()> for VmcallRawError {
+    fn from(_: ()) -> Self {
+        VmcallRawError::Illegal
+    }
+}
+
+impl error::Error for VmcallRawError {}
+
+impl From<VmcallRawError> for io::Error {
+    fn from(e: VmcallRawError) -> Self {
+        io::Error::new(io::ErrorKind::Other, e)
+    }
+}
+
+impl From<TdVmcallError> for VmcallRawError {
+    fn from(_err: TdVmcallError) -> VmcallRawError {
+        VmcallRawError::TdVmcallErr
+    }
+}
+
+#[derive(Copy, Clone, PartialEq, Eq, Default, PartialOrd, Ord)]
+pub struct VmcallRawAddr {
+    pub transport_context: u64,
+}
+
+impl VmcallRawAddr {
+    pub fn new(mid: u64) -> Self {
+        VmcallRawAddr {
+            transport_context: mid as u64,
+        }
+    }
+
+    pub fn transport_context(&self) -> u64 {
+        self.transport_context as u64
+    }
+
+    pub fn set_transport_context(&mut self, mid: u64) {
+        self.transport_context = mid as u64;
+    }
+}
+
+impl fmt::Display for VmcallRawAddr {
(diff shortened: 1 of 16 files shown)


jyao1:
@mjjagasi , please fix CI failure.

For DCO, please add "Signed-off-by: " tag.

src/devices/vmcall_raw/src/transport/vmcall.rs
```diff
@@ -0,0 +1,200 @@
...
+    })
+    .await
+}
+
+fn process_buffer(buffer: &mut [u8]) -> (&mut [u8], u32, u32) {
+    assert!(buffer.len() >= 8, "Buffer too small!");
+
+    let (int_bytes, remaining) = buffer.split_at_mut(8); // Split at 8th byte
+
+    let int1 = u32::from_le_bytes(int_bytes[0..4].try_into().unwrap()); // First 4 bytes
```
jyao1, line 154: we should give a meaningful field name, instead of int1, int2.

src/migtd/src/migration/session.rs
```diff
@@ -215,6 +314,58 @@ pub fn shutdown() -> Result<()> {
     Ok(())
 }
 
+fn process_buffer(buffer: &mut [u8]) -> (u32, u32) {
+    let (int_bytes, _remaining) = buffer.split_at_mut(8); // Split at 8th byte
+
+    let int1 = u32::from_le_bytes(int_bytes[0..4].try_into().unwrap()); // First 4 bytes
+    let int2 = u32::from_le_bytes(int_bytes[4..8].try_into().unwrap()); // Next 4 bytes
```
jyao1, line 321: we should give a meaningful field name, instead of int1, int2.

src/devices/vmcall_raw/src/transport/vmcall.rs
```diff
@@ -0,0 +1,200 @@
...
+            return Poll::Pending;
+        }
+
+        Poll::Ready(Ok(data_length as usize))
+    })
+    .await
+}
+
+fn process_buffer(buffer: &mut [u8]) -> (&mut [u8], u32, u32) {
+    assert!(buffer.len() >= 8, "Buffer too small!");
```
jyao1, line 150: how do you guarantee buffer.len() >= 8 ?

src/migtd/src/migration/session.rs
```diff
@@ -215,6 +314,58 @@ pub fn shutdown() -> Result<()> {
     Ok(())
 }
 
+fn process_buffer(buffer: &mut [u8]) -> (u32, u32) {
+    let (int_bytes, _remaining) = buffer.split_at_mut(8); // Split at 8th byte
```
jyao1, line 318: how do you guarantee buffer.len() >= 8 ?

src/devices/vmcall_raw/src/transport/vmcall.rs
mjjagasi: done! please review the latest

src/migtd/src/migration/session.rs
mjjagasi: done! please review the latest

src/migtd/src/migration/session.rs
mjjagasi: Windows PoC stack changes are made to ensure that minimum buffer length is 8 bytes for the first fragment

src/devices/vmcall_raw/src/transport/vmcall.rs
mjjagasi: Windows PoC stack changes are made to ensure that minimum buffer length is 8 bytes for the first fragment

jyao1: This is NOT documented. Even if it is documented, a bad VMM might not follow the spec.

MigTD need to handle this case. Using `assert` is NOT the right way.

src/migtd/src/migration/session.rs
```diff
@@ -121,79 +127,174 @@ pub fn query() -> Result<()> {
...
+                mig_td_cid: 0,
+                mig_channel_port: 0xAC9,
+                quote_service_port: 0,
+            };
+
+            if wfr.migration_source == 1 {
+                mig_socket_info.mig_td_cid = 33;
+            } else {
+                mig_socket_info.mig_td_cid = 34;
+            }
```
jyao1, line 201: Why we still need fill mig_socket_info for vmcall_raw ?

I believe it should NOT be required.

src/migtd/src/migration/session.rs
```diff
@@ -121,79 +127,174 @@ pub fn query() -> Result<()> {
...
+            };
+
+            for i in 0..4 {
+                mig_info.target_td_uuid[i] = wfr.target_td_uuid[i];
+            }
+
+            let mut mig_socket_info = MigtdStreamSocketInfo {
+                communication_id: 0,
+                mig_td_cid: 0,
+                mig_channel_port: 0xAC9,
```
jyao1, line 193: What is 0xAC9? Why we need this info?

jyao1: Can we set `mig_socket_info` to None in vmcall_raw?

src/migtd/src/migration/session.rs
mjjagasi: Removed, made mig_socket_info (MigtdStreamSocketInfo) as Option in data.rs

src/migtd/src/migration/session.rs
mjjagasi: Yes, done!

jyao1:
When you fix the CI issue, could you please split the td-shim upgrade to a standalone commit?
That will help us narrow down the potential issue.

src/migtd/src/bin/migtd/main.rs
```diff
@@ -142,7 +142,17 @@ fn handle_pre_mig() {
                         .await
                         .map(|_| MigrationResult::Success)
                         .unwrap_or_else(|e| e);
-                    let _ = report_status(status as u8, request.mig_info.mig_request_id);
+
```
gaojiaqi7, line 145: Can these two code blocks be merged? vmcall-raw is a feature used to distinguish GHCI versions and should not be used to differentiate whether report status is async/await.

gaojiaqi7 (changes requested):

src/devices/vmcall_raw/src/transport/vmcall.rs
```diff
@@ -0,0 +1,197 @@
...
```
gaojiaqi7, line 147: It will be safer to return an Option or Result for this function. If the buffer is too small, it returns a None or Error.

gaojiaqi7 (changes requested):

src/devices/vmcall_raw/src/lib.rs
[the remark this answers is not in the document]
mjjagasi: VmcallRaw is the name of the new transport. It only contains request ID field encapsulated in VmcallRawAddr to be flexible to enable adding new fields in future. Note: all other fields from vsock have been already removed (viz. port ID, rx_cnt, tx_cnt, state)

I have retained the function names -connect, shutdown, reset, send and receive and each has a reduced functionality vis-à-vis the original vsock implementations -I found these function names relevant even in the new VmcallRaw as well.

connect() -is used to setup the receive queue (BtreeMap) and register the interrupt callback. The Request/Response messages have been removed.
shutdown() -is used in the main flow so I retained it. The shutdown/reset messages have been removed. I wanted to keep the main interface similar to other transports.
reset() -removes the packet queue for this request ID and also removes the interrupt callback for it

1.	I can combine shutdown and reset in to 1 function and rename it as finish or complete -is that okay?
2.	I can rename connect to initialize -is that okay?

vmcall_raw_transport_enqueue and vmcall_raw_transport_dequeue are intermediate functions that in turn invoke the td-shim calls for the tdvmcalls. I believe they’re necessary or do you want to move them to send/recv itself?

All vsock headers have been removed.

src/migtd/src/bin/migtd/main.rs
mjjagasi: For report status and wait for request, have done different implementations. I am not sure if I should disturb the original function because it may deal with other transports that I am not familiar with. Can you/Jiewen make this change to the original function(s) called by other transports?

jyao1 (approved):

PR #414, opened 2025-07-16
Support engine-svn mapping and signing

gaojiaqi7:


src/policy/src/v2/servtd_collateral.rs
```diff
@@ -0,0 +1,236 @@
...
+            rtmr0: bytes_to_hex_string(
+                report
+                    .get_migtd_info_property(&MigTdInfoProperty::Rtmr0)
+                    .ok()?,
+            ),
+            rtmr1: bytes_to_hex_string(
+                report
+                    .get_migtd_info_property(&MigTdInfoProperty::Rtmr1)
+                    .ok()?,
+            ),
```
jyao1, line 178: For completeness, we should get RTMR2, RTMR3 here.

But RTMR2 and RTMR3 can be skipped if the mapping file does not has those values.

PR #457, opened 2025-09-22
tools: introduce servtd-collateral-generator and json-signer / extend migtd-policy-generator

gaojiaqi7:


jyao1 (approved):

jyao1:
Approve for the tool.
We need to update readme later, to include a full step-by-step flow.

PR #489, opened 2025-10-20
Update collaterals implementation

gaojiaqi7:
- Optimize the size of collaterals, extract duplicate contents such as Root CA CRL, QE identity, PCK CRL and their issuer chains.
 - Add SGX Root CA in collaterals


tools/migtd-collateral-generator/src/collateral.rs
```diff
@@ -2,94 +2,102 @@
...
         let qe_identity = String::from_utf8(qe_identity)
             .map_err(|e| anyhow!("Invalid UTF-8 in QE identity: {}", e))?;
 
-        Ok(Collateral {
+        Ok(Collaterals {
             major_version: 1,
             minor_version: 0,
             tee_type: 0x81,
             pck_crl_issuer_chain,
+            root_ca,
```
jyao1: `root_ca` should be before `pck_crl_issuer_chain`.

jyao1 (approved):

PR #490, opened 2025-10-20
deps: update linux-sgx submodule to `migtd_2.26`

gaojiaqi7:


jyao1 (approved):

haitaohuang:
Hi Just FYI to sync understanding here. Not sure if this is related to the build error for Integrated Test on TDX. On my side, when I test with our AzCVMEmu mode (using azure EMR based VM with TCP for comm), I notice the verify_quote_integrity() in  the migtd_2.26 branch returns 774 bytes for verified report size, which breaks v1 policy check.  My fix was to change migtd to expect larger report size and use the known old value as minimal:
https://github.com/haitaohuang/MigTD/commit/41d00cd66ebad364c60cee89ef4923fff03d6fe8

src/attestation/src/lib.rs
```diff
@@ -13,7 +13,7 @@ pub mod root_ca;
 
 pub use attest::*;
 
-pub const TD_VERIFIED_REPORT_SIZE: usize = 734;
+pub const TD_VERIFIED_REPORT_SIZE: usize = 774;
```
jyao1: how this is calculated?
Please provide comment or use sizeof, etc

src/policy/src/lib.rs
```diff
@@ -25,7 +25,7 @@ use td_shim::event_log::{
     PLATFORM_CONFIG_SECURE_AUTHORITY, PLATFORM_CONFIG_SVN, PLATFORM_FIRMWARE_BLOB2_PAYLOAD,
 };
 
-pub const REPORT_DATA_SIZE: usize = 734;
+pub const REPORT_DATA_SIZE: usize = 774;
```
haitaohuang: this breaks get_tcb_date_and_status_from_suppl_data where this value is expected to be 734

PR #506, opened 2025-10-24
Fix CI SPDM test build to build test bin with spdm_attestation feature

IntelCaisui:


sh_script/build_final.sh
```diff
@@ -326,6 +327,9 @@ function enroll() {
 
 ./sh_script/preparation.sh
 
+echo "The spdm config is $SPDM_CONFIG"
```
IntelCaisui: Containing test output to check CI build environment, will remove it with further update of this pr.

PR #523, opened 2025-10-28
Fix SPDM_CONFIG env. variable setting in build command

bodzhang:
- Set SPDM_CONFIG env. variable in xtask build.rs for cargo image to use
config/spdm_config.json instead of the default spdmlib config.

Signed-off-by: Bo Zhang (ACC) <zhanb@microsoft.com>


jyao1:
patch is good.
But CI failed. Would you please fix that?

PR #531, opened 2025-10-30
spdm add const defs for vdm messages

IntelCaisui:


jyao1:
please resolve the conflict.

PR #547, opened 2025-11-09
Add get_quote implementation for igvm.

agokarn:
This change implements IGVM-based attestation quote retrieval for MigTD, providing a Rust implementation for obtaining TDX quotes through a Service TD (ServTD) interface when running in an IGVM (Isolated Guest Virtual Machine) environment. The implementation is specifically designed for Azure's virtualization stack, enabling secure communication with the VMM via GHCI to request and receive attestation quotes. This functionality is conditionally compiled with the igvm-attest feature flag, making it part of Azure's IGVM-specific deployment of TDX confidential computing capabilities.

### DIFF
--- a/src/attestation/src/igvmattest.rs
+++ b/src/attestation/src/igvmattest.rs
@@ -1,0 +1,157 @@
+// Copyright (c) 2022 Intel Corporation
+//
+// SPDX-License-Identifier: BSD-2-Clause-Patent
+
+use crate::{attest::TD_QUOTE_SIZE, Error};
+#[cfg(feature = "igvm-attest")]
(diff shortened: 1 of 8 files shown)


src/attestation/src/igvmattest.rs
```diff
@@ -0,0 +1,152 @@
+use crate::{attest::TD_QUOTE_SIZE, Error};
```
jyao1: please add license header.

jyao1:
Please don't use a dedicated commit to address the comment. Please squash them, by using force push.

The commit split should be based on the feature, but not how you develop the code.
Each commit should be a complete commit, and to address one specific existing problem.

PR #597, opened 2025-11-20
Refactor code in main.rs

agokarn:
The pull request refactors code in Main.rs. Split do_measurements() and get_policy_and_measure into two variants – for policy_v2 and existing v1 policy.


agokarn:
> Please split the PR to 2 different PRs:
> 
> 1. Modify Makefile for Azure
> 2. refactor code in main.rs
> 
> They are totally different tasks to resolve different problem.
> 
> I might approve 1) Azure specific change quickly. But I need deep understanding on 2), what problem you want to resolve by refactoring.
> 
> **Grand rule: each PR resolves and only resolves one specific problem.**

jyao1:
Why we need change sh_script/fuzzing.sh?

The commit message is not right. It still has the old content.

agokarn:
> The pull request refactors code in Main.rs. Split do_measurements() and get_policy_and_measure into two variants – for policy_v2 and existing v1 policy.



> Why we need change sh_script/fuzzing.sh?
> 
> The commit message is not right. It still has the old content.

I can close this and start a new PR>

PR #607, opened 2025-11-20
disable serial logs by default for release binary

mgudaram:


src/migtd/Cargo.toml
```diff
@@ -23,7 +23,7 @@ chrono = { version = "0.4", default-features = false, features = ["alloc"] }
 crypto = { path = "../crypto" }
 futures-util = { version = "0.3.17", default-features = false }
 lazy_static = { version = "1.0", features = ["spin_no_std"] }
-log = { version = "0.4.13" }
+log = { version = "0.4.13", features = ["release_max_level_off"] }
```
agokarn: Will this impact entry log later on when the that uses the td_logger crate later on?

mgudaram: At this moment I am not sure, it would depend on how we implement it (how we create the switch to decide serial vs memory). if we want to, we can honor it for entrylog scenario i.e. serial logs disabled.

PR #622, opened 2025-11-21
migtd: refactor pre-session data implementation to align with spec

gaojiaqi7:
Closes https://github.com/intel/MigTD/issues/621

### DIFF
--- a/src/migtd/src/migration/session.rs
+++ b/src/migtd/src/migration/session.rs
@@ -632,16 +632,17 @@ pub fn report_status(status: u8, request_id: u64) -> Result<()> {
 
 #[cfg(feature = "policy_v2")]
 #[repr(C)]
-struct MigtdMessageHeader {
+struct PreSessionMessage {
     pub r#type: u8,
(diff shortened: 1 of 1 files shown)


src/migtd/src/migration/session.rs
```diff
@@ -692,79 +731,133 @@ async fn send_policy<T: AsyncRead + AsyncWrite + Unpin>(
...
-    while recvd < header_buffer.len() {
-        let n = transport
-            .read(&mut header_buffer[recvd..])
-            .await
-            .map_err(|_| MigrationResult::NetworkError)?;
-        recvd += n;
+    let command = PreSessionMessage::read_from_bytes(&header_buffer)
+        .ok_or(MigrationResult::InvalidParameter)?;
+    if command.r#type != PreSessionMessage::START_SESSION_TYPE {
+        return Err(MigrationResult::InvalidParameter);
```
liuw1: The length of start session must be 0, it's better to add check here.

PR #624, opened 2025-11-21
Update td-shim to the latest upstream

haitaohuang:
To pull in changes for option to unmeasure CFV for igvm builds


haitaohuang:
make it draft to test toolchain 1.88 as needed by td-shim

jyao1 (approved):

PR #635, opened 2025-11-24
Add VmmLoggerBackend

agokarn:
This pull-request is related to a new feature extension to existing VMM logging mechanism (connecting it to RUST log crate) and also resolves the request: [https://github.com/https://github.com/intel/MigTD/issues/571]

Instead of using entrylog wherever logs need to be added to the vmm log area, the existing logging macros log::info!, log::error!, etc can be routed to the vmm log area. 

1. init_vmm_logger will call log::set_logger, which is a global, one-time initialization function from the log crate that installs a logger implementation as the destination for all subsequent log!, info!, debug!, error!, etc. macros in the entire program.

2. VmmLoggerBackend will implement the Log trait, which has the following three functions -

       impl Log for VmmLoggerBackend {
              fn enabled(&self, metadata: &Metadata) -> bool {
              }

              fn log(&self, record: &Record) {
              }

              fn flush(&self) {}
       }
       
This is similar to what is implemented in td_logger. When info level is enabled and log::info! is called, it invokes the log function. The implementation in this function in VmmLoggerBackend does two things -

-  Calls entrylog(&msg.into_bytes(), record.level(), request_id);
    The last parameter here is request_id, which is got from let request_id = CURRENT_REQUEST_ID.load(Ordering::SeqCst); and is set by the function set_current_request_id.

- Calls td_logger::dbg_write_string(&format!("{} - {}\n", record.level(), record.args())); for backward compatibility to continue sending output to the serial port if it is not blocked.

### DIFF
--- a/src/migtd/src/migration/logging.rs
+++ b/src/migtd/src/migration/logging.rs
@@ -11,7 +11,8 @@ use alloc::format;
 use alloc::vec::Vec;
 use core::sync::atomic::{AtomicBool, AtomicU32, AtomicU64, AtomicU8, Ordering};
 use lazy_static::lazy_static;
-use log::Level;
+use log::kv::{Error, Key, Value, VisitSource};
+use log::{Level, LevelFilter, Metadata, Record, SetLoggerError};
 #[cfg(not(test))]
 use raw_cpuid::CpuId;
 use spin::Mutex;
@@ -23,6 +24,7 @@ use zerocopy::{transmute_ref, AsBytes, FromBytes, FromZeroes};
 const PAGE_SIZE: usize = 0x1_000;
 #[cfg(not(test))]
 const TDCALL_STATUS_SUCCESS: u64 = 0;
+const MIGRATION_REQUEST_ID_SENTINEL: u64 = 0xFFFF_FFFF_FFFF_FFFF;
 
 type Result<T> = core::result::Result<T, MigrationResult>;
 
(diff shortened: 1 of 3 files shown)


src/migtd/src/migration/logging.rs
```diff
@@ -1011,4 +1056,80 @@ mod test {
...
+        assert!(result.is_ok());
+
+        let result = enable_logarea(log_max_level, 0, &mut data).await;
+        assert!(result.is_ok());
+
+        // Set request ID before logging
+        set_current_request_id(12345);
+
+        // call log::info macro
+        log::info!("Test message");
```
agokarn: Notice this is entrylog in other tests. Here log::info!("Test message") will show up in the VMM log area as well.

src/migtd/src/migration/logging.rs
```diff
@@ -471,6 +472,50 @@ pub fn entrylog(msg: &Vec<u8>, loglevel: Level, request_id: u64) {
...
 }
 
+lazy_static! {
+    static ref CURRENT_REQUEST_ID: AtomicU64 = AtomicU64::new(0);
+}
+
+/// Set the current request ID for logging context
+pub fn set_current_request_id(request_id: u64) {
+    CURRENT_REQUEST_ID.store(request_id, Ordering::SeqCst);
+}
```
jyao1: @mgudaram, would you please comment this?
How does current code work without this API?

src/migtd/src/migration/logging.rs
```diff
@@ -471,6 +472,50 @@ pub fn entrylog(msg: &Vec<u8>, loglevel: Level, request_id: u64) {
     }
 }
 
+lazy_static! {
+    static ref CURRENT_REQUEST_ID: AtomicU64 = AtomicU64::new(0);
```
mgudaram: Minor change to set default value as u64::MAX. If logs are added in the area MigRequestID is not available, then it takes "0" as default which is valid and could lead to misinterpretation. Moving forward, probably this value should be reset to default at places applicable, if not it could point to previous RequestID at places requestID is not available.

_According to GHCI spec, "RequestID: The ID of the migration request associated with the log entry, if any. If not associated with any MigRequestId, the sentinel value 0xFFFFFFFFFFFFFFFF."_

src/migtd/src/migration/logging.rs
agokarn: It works because each call to entrylog takes in request_id as a parameter. If we want to call log::info and translate that to a entrylog call, the request_id has to be present when the call is made. One way is to set this value with a new API and qurey for it from log function. @mgudaram can confirm

mgudaram: > @mgudaram, would you please comment this? How does current code work without this API?

Right, as arthi mentioned, I use Request_ID as an argument to the function, so in MultiVP/other scenarios, it always has the intended Request_ID to function call or a Default sentinal value.

jyao1:
I am OK with this change.

@mgudaram , could you please also take a look?

jyao1 (approved):

PR #649, opened 2025-11-27
config: update v1 policy

gaojiaqi7:
- Align the policy key order with the spec.
    
- Update `sgxtcbcomponents`, `tdxtcbcomponents` and `TDXModuleSVN`.


jyao1:
Please explain the problem statement. Why we need this change?

Issue #650, opened 2025-11-28
Convert `migtd-policy-generator` and `migtd-collateral-generator` to async


gaojiaqi7:
Currently the `migtd-policy-generator` and takes about 20 - 30 seconds to fetch all required data from PCS server. It takes so long because it send the HTTP requests sequentially; handling them asynchronously will save a lot of time.

PR #651, opened 2025-11-28
Speed up the `migtd-collateral-generator` and `migtd-policy-generator`

gaojiaqi7:
Convert the http requests to async mode. This saves roughly 90% of the runtime—the measurement dropped from 20–30 seconds down to about 2–3 seconds.

Changes:

 - Replace `curl` with a async http client library `reqwest`
 - Use `tokio` as async runtime.
 - Convert all the PCS server access API to async

Closes https://github.com/intel/MigTD/issues/650

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -1386,6 +1386,12 @@ dependencies = [
 ]
 
 [[package]]
+name = "mime"
+version = "0.3.17"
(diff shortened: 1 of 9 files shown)


jyao1:
can we add to CI?

PR #654, opened 2025-12-01
tools: add migtd-policy-verifier

haitaohuang:


haitaohuang:
This tool can be used in pipeline to sanity check signed policy before embedding it into CFV within a MigTD for testing

jyao1 (approved):

PR #659, opened 2025-12-02
CI: enable test for tools

gaojiaqi7:


jyao1 (dismissed):

tools/migtd-collateral-generator/src/pcs_client.rs
```diff
@@ -269,7 +269,9 @@ mod tests {
     ///
     /// Run with: cargo test test_thim_headers -- --nocapture
     /// Set AZURE_REGION environment variable or it defaults to "useast"
+    /// Ignored in CI because it requires network access to Azure THIM.
```
jyao1: @bodzhang, do you know why we cannot access Azure THIM?

PR #679, opened 2025-12-18
Avoid use vdm payload struct with slices

IntelCaisui:
To reduce the stack size consumed by avoid allocating large slices
  caused by vdm structs have includes vdm payload slices.


IntelCaisui:
Not Ready, need spdm update.

PR #689, opened 2026-01-18
Add initial spdm attestation support for rebinding

IntelCaisui:
Add Spdm attestation support for rebinding flow.


config/spdm_config.json
```diff
@@ -17,7 +17,7 @@
         "sender_buffer_size": 4096,
         "receiver_buffer_size": 4096
     },
-    "max_spdm_msg_size": 65536,
+    "max_spdm_msg_size": 131072,
```
IntelCaisui: Will revise it.

jyao1: Please revert if it is NOT tested.

jyao1 (approved):

PR #701, opened 2026-02-04
Fix Migtd attestation doesn't close transport

IntelCaisui:


src/migtd/src/migration/session.rs
```diff
@@ -951,6 +953,10 @@ async fn migration_src_exchange_msk(
         e
     })?;
     log::info!("MSK exchange completed\n");
+
+    let mut transport_lock = device_io_ref.lock();
+    let transport = transport_lock.deref_mut();
+    shutdown_transport(&mut transport.transport, info.mig_info.mig_request_id).await?;
```
jyao1: I don't think we should call shutdown_transport() in migrate_xxx_exchange_msk().

shutdown_transport() should be called by uplevel after migrate_xxx_exchange_msk().

PR #733, opened 2026-02-26
Sanitize session logging.

agokarn:
An internal tool has flagged this code. Even though the security reviewer doesn't think it exploitable, it violates the tool and hence have requested the fix.


haitaohuang (approved):
looks good to me

jyao1:
please add description - why this change is needed.

jyao1 (approved):

PR #735, opened 2026-03-03
MigTD: handle target TD without SERVTD_EXT

haitaohuang:
On platforms with rebind support but no support or TD opts out for SERVTD_EXT, TDG.servtd.rd on the SERVTD_EXT fields in TDCS of a target TD would return zeros. This change reads TDCS.ATTRIBUTES to check bit 17 and makes SERVTD_EXT optional throughout the rebinding flow:

- read_servtd_ext() reads TDCS.ATTRIBUTES via tdcall_servtd_rd and returns None when SERVTDEXT bit is not set
- write_approved_servtd_ext_hash() accepts Option and is a no-op when None
- Certificate generation/verification: servtd_ext extension is conditionally included and tolerated when missing
- Policy verification (authenticate_rebinding_old) skips init report verification entirely when servtd_ext is unavailable, since the init TD report cannot be verified without servtd_info_hash
- SPDM VDM messages send zero-length servtd_ext element when not available; receiver handles it gracefully
- All TDCS write operations (write_servtd_rebind_attr, write_approved_servtd_ext_hash) are skipped when servtd_ext is None

### DIFF
--- a/src/migtd/src/mig_policy.rs
+++ b/src/migtd/src/mig_policy.rs
@@ -270,7 +270,7 @@ mod v2 {
         init_policy: &[u8],
         init_event_log: &[u8],
         init_td_report: &[u8],
-        servtd_ext_src: &[u8],
+        servtd_ext_src: Option<&[u8]>,
     ) -> Result<Vec<u8>, PolicyError> {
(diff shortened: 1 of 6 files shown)


gaojiaqi7 (commented):
Hi @haitaohuang , do you think we should treat it an error when SERVTDEXT isn't supported, and report the status accordingly?

PR #742, opened 2026-03-11
changes to support provisional logs

mgudaram:
Provisional logs store logs until VMM enables memory logging and copies these logs into the memory area. Provisional logs would be 4KB circular buffer per vCPU.


jyao1:
I am OK with this change. May I know what kind of test you have run to ensure the code is correct?

jyao1 (approved):

PR #769, opened 2026-04-02
Fix migtd layout configuration

MichalTarnacki:


sgrams:
I think we should work to resolve this on TD-Shim end.

PR #814, opened 2026-04-27
fix(migtd): Remove cached LOCAL_TCB_INFO

haitaohuang:
Currently, migtd requests a quote and cache a static tcb during MigTD startup. This adds extra latency for migtd to start and the cached tcb would become invalid after an impactless update for FW.

The only benefit of doing it is to verify the policy and fails fast if the policy is malformed. That can be done by build pipelines offline.

Remove the LOCAL_TCB_INFO static cache and init_tcb_info() function. get_local_tcb_evaluation_info() now generates a fresh quote and computes evaluation data on every call.

### DIFF
--- a/src/migtd/src/bin/migtd/main.rs
+++ b/src/migtd/src/bin/migtd/main.rs
@@ -399,15 +399,6 @@ fn initialize_policy() -> String {
         );
     });
 
-    // Initialize and verify the migration policy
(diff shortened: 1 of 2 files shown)


jyao1:
@haitaohuang please resolve the conflict.

jyao1 (approved):

PR #817, opened 2026-04-28
virtio-serial driver bugfixes

Martinits:
Fix 2 migtd virtio-serial driver bugs:
- Missing control msgs since qemu may put multiple control msg but the driver pops one each time. 
- A late PORT_OPEN event may cause a PortNotAvailable when port.open(), try to drain control queue when PortNotAvailable and retry open.

### DIFF
--- a/src/devices/virtio_serial/src/lib.rs
+++ b/src/devices/virtio_serial/src/lib.rs
@@ -482,7 +482,7 @@ impl VirtioSerial {
 
         self.timer.reset_timeout();
 
-        if self
+        while self
             .queues
             .index(CONTROL_RECEIVEQ as usize)
             .borrow_mut()
@@ -537,7 +537,6 @@ impl VirtioSerial {
             ControlEvent::DeviceAdd => {
                 self.fill_port_queue(port_id)?;
                 self.port_ready(port_id)?;
-                self.recv_control()?;
             }
             ControlEvent::DeviceRemove => {
                 self.connected_ports.remove(&port_id);
--- a/src/devices/virtio_serial/src/lib.rs
+++ b/src/devices/virtio_serial/src/lib.rs
@@ -467,6 +467,53 @@ impl VirtioSerial {
         Ok(())
     }
 
+    /// Non-blocking control-queue pump: drains any already-pending control
+    /// messages without waiting for new IRQ/events.
+    fn recv_control_nonblock(&mut self) -> Result<()> {
+        let vq = self.queues.index(CONTROL_RECEIVEQ as usize);
+        if !vq.borrow_mut().can_pop() {
+            return Ok(());
+        }
+
+        while self
+            .queues
+            .index(CONTROL_RECEIVEQ as usize)
+            .borrow_mut()
+            .can_pop()
+        {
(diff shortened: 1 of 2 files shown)


jyao1:
could you please fix CI failure?

Martinits:
@jyao1 CI failures fixed.

sgrams:
Two small things to address before we can merge this:
1) `recv_control_nonblock` is missing the `dma_allocation.get -> safe_len = min(vq_buf.len, record.dma_size)` clamp that recv_control performs. Both paths consume a host-controlled vq_buf.len; the clamp is a deliberate defense against an overlong len. Worth mirroring so both helpers are equally hardened.
2) Retry-exhaustion path silently returns `Ok(port)` - if all 100 retries fail with PortNotAvailable, the for loop falls through and setup_transport returns an unopened port. The failure then surfaces at the first send/recv with no breadcrumb. Track success with a bool and return Err(VirtioSerialError::PortNotAvailable(VIRTIO_SERIAL_PORT_ID)) + a log::error! on exhaustion.

@Martinits can you please help to address it? It should be a 3-line change, see below:

For 1)
```
            for vq_buf in &h2g {
                if let Some(record) = self.dma_allocation.get(&vq_buf.addr) {
                    let safe_len = core::cmp::min(vq_buf.len as usize, record.dma_size);
                    let control_msg = unsafe {
                        core::slice::from_raw_parts(vq_buf.addr as *const u8, safe_len)
                    };
                    self.handle_control_msg(control_msg)?;
                    self.free_dma_memory(vq_buf.addr)
                        .ok_or(VirtioSerialError::OutOfResource)?;
                }
            }
```

For 2)
```
        let port = VirtioSerialPort::new(VIRTIO_SERIAL_PORT_ID);
        let mut opened = false;
        for _ in 0..VIRTIO_SERIAL_OPEN_RETRY_TIMES {
            match port.open() {
                Ok(()) => {
                    opened = true;
                    break;
                }
                Err(VirtioSerialError::PortNotAvailable(_)) => {
                    // `PORT_OPEN` can arrive after init_control; pump control queue during retry.
                    // If we drained pending control msgs, try open again immediately to avoid
                    // waiting for the next retry tick.
                    Timer::after(VIRTIO_SERIAL_OPEN_RETRY_DELAY).await;
                    VirtioSerial::try_poll_control()?;
                }
                Err(e) => return Err(e.into()),
            }
        }
        if !opened {
            log::error!(
                "virtio-serial port {} not available after {} retries ({}ms total)\n",
                VIRTIO_SERIAL_PORT_ID,
                VIRTIO_SERIAL_OPEN_RETRY_TIMES,
                VIRTIO_SERIAL_OPEN_RETRY_TIMES * VIRTIO_SERIAL_OPEN_RETRY_DELAY.as_millis() as usize,
            );
            return Err(VirtioSerialError::PortNotAvailable(VIRTIO_SERIAL_PORT_ID).into());
        }
        return Ok(port);
```

sgrams (changes requested):
See my previous comment

Martinits:
@sgrams Thanks, I've patched your suggestions.

sgrams (approved):

PR #820, opened 2026-05-05
fix(config): relocate TempMem sections to guest RAM

MichalTarnacki:
Add TempMemBase to image_layout.json and update metadata.json to place Mailbox, TempStack, and TempHeap at 0x7DF000-0x800000 (guest RAM) instead of the firmware pflash range (0xFF000000+).

Split PermMem into two regions (0x0-0x7DF000 and 0x800000-0x2000000) to avoid overlap with relocated TempMem sections.

This allows unpatched upstream QEMU to accept TempMem pages via E820 RAM entries during TD initialization, eliminating the 'Failed to accept memory for TDVF section' error.

### DIFF
--- a/config/image_layout.json
+++ b/config/image_layout.json
@@ -6,5 +6,6 @@
     "Metadata": "0x001000",
     "Payload": "0xEE6000",
     "Ipl": "0x50000",
-    "ResetVector": "0x008000"
+    "ResetVector": "0x008000",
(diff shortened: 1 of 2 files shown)


jyao1:
could you please resolve conflict?

sgrams:
@MichalTarnacki can you please help to resolve conflict here? Thanks

PR #827, opened 2026-05-06
fix(migtd): remove init_event_log from rebinding attestation

### LINKED ISSUE #818: init event log should be removed
src migtd event log is used as init event log here:
https://github.com/intel/MigTD/blob/main/src/migtd/src/spdm/spdm_req.rs#L1014

it was checked against init mig tdinfo which will fail.
We should remove the init event log per spec.

MichalTarnacki:
Per GHCI 1.5, MIGTD_DATA only contains tdinfo (type 0) and there is no init_event_log entry. The EventLogInit VDM element was sending the local event log as a substitute, which would fail verification against init tdinfo RTMRs since they belong to a different TD instance.

Remove EventLogInit from the VDM ExchangeRebindAttestInfo request/response, drop the init_event_log parameter from authenticate_rebinding_old(), and remove the corresponding X.509 extension from the RATLS certificate. The rebind attest request element count is reduced from 7 to 6.

Closes #818


haitaohuang (commented):
LGTM, thanks

sgrams (approved):

PR #839, opened 2026-05-20
fix(migtd): fix code coverage collection for fuzzing tests

### LINKED ISSUE #463: Code coverage collection for fuzzing test is not workable

MichalTarnacki:
Coverage collection via the -c flag was broken due to multiple issues:

- AFL: RUSTFLAGS="-C instrument-coverage" was applied during cargo afl build/fuzz, conflicting with AFL's own instrumentation and fork server. Coverage profraw files were either corrupt or missing. Fix by building a separate non-AFL instrumented binary (without --features fuzz) after fuzzing, and replaying the AFL queue through it to generate valid coverage data.

- AFL: grcov searched from "." which could pick up stale profdata files from other runs. Fix by writing profraw to a dedicated cov_profraw/ directory and pointing grcov only there.

- AFL: the queue replay glob (queue/*) included the .state directory, causing panics. Fix by using find -type f to iterate only regular files.

- Libfuzzer: grcov --binary-path pointed to fuzz/target/x86_64-unknown-linux-gnu/release/<target> (a file) but should be the release/ directory. Additionally, grcov searched from "." picking up incompatible profdata. Fix by pointing grcov to the specific fuzz/coverage/<target>/raw directory where cargo fuzz coverage writes profraw files.

- Libfuzzer run_all_case: inverted conditional (if [ ! -d ... ]) meant the old coverage directory was never cleaned up. Fix the logic.

Fixes #463

### DIFF
--- a/sh_script/fuzzing.sh
+++ b/sh_script/fuzzing.sh
@@ -139,11 +139,6 @@ run_single_case() {
     echo $test_case | grep "^afl"
     if [ "$?" == "0" ];then
         echo "The test method is afl"
-        if [ "${collect_coverage}" == "YES" ]; then
-            export RUSTFLAGS="-C instrument-coverage"
-            export LLVM_PROFILE_FILE="fuzz-%p-%m.profraw"
-            find . -name "*.profraw" | xargs rm -rf
-        fi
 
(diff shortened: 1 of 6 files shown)


jyao1 (approved):

jyao1:
what is current coverage data?

MichalTarnacki:
image

PR #857, opened 2026-05-27
fix(servtd_ext): remove broken init_attr == curr_attr check

### LINKED ISSUE #845: Compare init_attr==curr_attr not needed?
#831 seem to be referring old version of spec when we have MigTDData defined. Now we only have init_tdinfo from VMM.

sgrams:
The verify_servtd_attr and read_servtd_ext functions compared CURR_SERVTD_ATTR against INIT_ATTR read from the target TD. This check is invalid because INIT_ATTR is only populated when the rebind feature is supported, while the initial binding does not set it.

Remove the redundant INIT_ATTR comparison and keep only the hardcoded EXPECTED_SERVTD_ATTR verification which ensures the VMM wrote the correct value.

Closes #845


jyao1:
@haitaohuang , could you please take a look?

haitaohuang:
LGTM, thanks

jyao1 (approved):

PR #860, opened 2026-05-27
feat(vmcall_raw): handle VMM cancellation (0x302)

haitaohuang:
The VMM signals migration session cancellation by writing data_status byte[0]=0x02, byte[1]=0x03 (0x302) to the shared buffer.

Previously, 0x302 was treated as a generic TdVmcallErr, reported as SecureSessionError (6), change to propagate and report it as VmmCanceled (10) in ReportStatus.

In the SPDM path, MigtdTransport's send/receive implement the SpdmDeviceIo trait which cannot represent this error in its return types (SpdmResult and Result<usize, usize>). Log a warning when ConnectionAborted is detected so the VMM cancellation is visible in logs, even though the typed error cannot propagate through the SPDM library's trait boundary.

### DIFF
--- a/src/devices/vmcall_raw/src/transport/vmcall.rs
+++ b/src/devices/vmcall_raw/src/transport/vmcall.rs
@@ -25,7 +25,9 @@ lazy_static! {
     pub static ref VMCALL_MIG_CONTEXT_FLAGS: Mutex<BTreeMap<u64, AtomicBool>> =
         Mutex::new(BTreeMap::new());
 }
-const TDX_VMCALL_VMM_SUCCESS: u8 = 1;
+const TDX_VMCALL_VMM_SUCCESS: u8 = 0x01;
+const TDX_VMCALL_STATUS_NOT_COMPLETED: u8 = 0x02;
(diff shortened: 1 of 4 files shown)


src/devices/vmcall_raw/src/transport/vmcall.rs
```diff
@@ -139,6 +139,15 @@ async fn vmcall_service_migtd_send(
         let data_status_bytes = data_status.to_le_bytes();
 
         if data_status_bytes[0] != TDX_VMCALL_VMM_SUCCESS {
+            // byte[0]=0x02 (not completed), byte[1]=0x03 (VMM cancel)
+            if data_status_bytes[0] == 0x02 && data_status_bytes[1] == 0x03 {
```
jyao1: can we define a MACRO/CONST value for 0x02 and 0x03?

haitaohuang: done. accidentally pushed wrong branch below, fixed too

jyao1 (approved):

PR #874, opened 2026-06-03
fix(mig_policy): drop init-ref policy evaluation

### LINKED ISSUE #848: evaluation of policy relative to init MigTD not needed?
https://github.com/intel/MigTD/blob/main/src/migtd/src/mig_policy.rs#L292
We are using init migtg tcb as relative reference and evaluate the old of rebinding on the new side using the new policy. Is this needed?

haitaohuang:
The destination's current mig_policy was being applied retroactively to the historical init→source transition. Remove as it's not needed.

fixes #848

### DIFF
--- a/src/migtd/src/mig_policy.rs
+++ b/src/migtd/src/mig_policy.rs
@@ -289,14 +289,6 @@ mod v2 {
             ))
             .ok_or(PolicyError::SvnMismatch)?;
 
-        // Use local policy's tcb_mapping with init tdinfo measurements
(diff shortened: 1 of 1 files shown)


haitaohuang:
please retrigger failed ci tests. failures seem to be network/ci platform issue

agokarn:
Changes look good

sgrams:
I restarted the jobs

sgrams (approved):

PR #878, opened 2026-06-04
security(spdm): bind rebind attestation to SPDM session TH1

haitaohuang:
Add verify_tdreport_data_binding() to check rebind TDREPORT REPORTDATA equals SHA384(prefix || TH1), preventing replay of attestation from a previous SPDM session. Applied on both requester and responder sides.

### DIFF
--- a/src/migtd/src/spdm/mod.rs
+++ b/src/migtd/src/spdm/mod.rs
@@ -223,6 +223,39 @@ pub fn verify_report_data_binding(
     verify_peer_report_data(supplemental_data, &report_data)
 }
 
+/// Verify that a TDREPORT's REPORTDATA is bound to the expected prefix and TH1.
+///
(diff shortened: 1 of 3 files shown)


agokarn (approved):

sgrams (approved):

sgrams:
Verified against MigTD Design Guide §4 (Mutual Authentication): TH1 must be bound into the TDREPORT, and the session must be torn down on auth failure - both satisfied here.
The prefix swap ("MigTDReq"/"MigTDRsp") and TH1 source (session.get_th1()) match the existing migration-path check in verify_report_data_binding, so the rebind path now has the same anti-replay/domain-separation guarantees the migration path already had. gen_tdreport at the generation sites already hashes prefix||TH1 into REPORTDATA via SHA-384, which lines up exactly with what the new verifier recomputes.

PR #869, opened 2026-05-29
 refactor(migtd): unify migration and rebinding VMM request buffer layout

haitaohuang:
Fold init_td_info into MigtdMigrationInformation
    
    Add init_td_info [u8; 512] field to MigtdMigrationInformation so both
    migration and rebinding use the same VMM request buffer layout. Delete
    RebindingInfo, InitData, and init_data.rs. Widen policy_v2 cfg gates
    that were incorrectly restricted to vmcall-raw only.
    
    Peer-to-peer wire protocol is unchanged.

fixes #868

PR #882, opened 2026-06-04
fix(rebinding): validate all reserved bytes 10..15

haitaohuang:
This was a regression introduced in 23a01d07f


agokarn:
Changes look good

haitaohuang:
Note: if https://github.com/intel/MigTD/pull/869 is merged, this is not needed

sgrams (approved):

PR #889, opened 2026-06-09
WIP: refactor(migtd/migration): drop Service.Query leaf and its tests

sgrams:
GHCI 1.5 v6 has no Service.Query sub-function (Tables 3-42 / 3-45 / 3-52 / 3-59 / 3-62 — only the <MigTD.*> per-API leaves: WaitForRequest, ReportStatus, Send, Receive). Service.Query was a v5 holdover used to probe VMM support before any real work; in v6 the support check is implicit in calling MigTD.WaitForRequest itself.

Remove:
  * migration::session::query() — the only caller of the Service.Query envelope path
  * ServiceQueryResponse struct and QUERY_COMMAND constant in migration::data
  * tests/test-td-payload's testservice case and its tcs002 entry in the JSON config
  * src/migtd/fuzz service_response and afl_service_response targets (and their seed corpus) which exclusively exercised the dead response wrapper


sgrams:
Deprecation of Service is not required for June release, old APIs are permittable

sgrams:
Closing for now

PR #887, opened 2026-06-09
add init tdinfo cross check

haitaohuang:
fixes #823
Currently rebased on #869

PR #890, opened 2026-06-09
 Add peer policy signing cert chain validation + key rotation support

haitaohuang:
Support policy signing certificate key rotation during migration, gated on the policy_v2 feature.

Wire format: Peer-data blob [u32 policy_len][policy][u32 chain_len][issuer_chain] is exchanged in pre-session and hashed into the EXTNID_MIGTD_POLICY_HASH X.509 extension. Replaces per-side rebinding wrappers with a single peer_data: Vec<u8> plumbed through pre_session_data, rebinding, ratls, and spdm.

Validation (validate_peer_cert_chain):

 - Root CA DER must match local chain
 - Leaf Subject Name must match
 - Every issuer in peer_chain[1..] must have BasicConstraints cA=TRUE (RFC 5280 Â§4.2.1.9) â blocks a stolen-leaf-key attack that forges [fake_leaf, legit_leaf, â¦]
 - Same checks applied to servtd_identity and servtd_tcb_mapping chains embedded in peer policy

CI / tooling:

 - key_gen.sh emits proper CA-bit intermediates; test fixtures regenerated, plus negative attacker_chain fixture
 - build_AzCVMEmu_policy_and_test.sh produces _a/_b/_pm_b/_pmi_b signed-policy variants rotating policy, tcb_mapping, and td_identity leaves independently
 - migtdemu.sh gains --src-* / --dst-* policy & chain flags
 - integration-emu.yml adds 3 key-rotation matrix entries + 2 multi-chain rotation tests

Tests: cargo test -p crypto --lib 10 passed; cargo test -p policy --lib 26 (+15 with policy_v2).

Currently based on #887


haitaohuang:
rebased

src/crypto/src/lib.rs
```diff
@@ -260,6 +263,55 @@ fn verify_signature_with_algorithm(
     }
 }
 
+/// Validates a peer's certificate chain against the local certificate chain.
+///
+/// Performs the following checks:
+/// 1. Verifies the peer chain's internal signature integrity
+/// 2. Root CA must match between local and peer chains
+/// 3. Leaf certificate Subject Name must match
```
jyao1, line 271: question: Only Root-CA match? What about intermediate certs?
Should we consider intermediate cert match as well? And just rotate the leaf cert?

Whatever the design is, then please document intermediate cert here.

haitaohuang: yeah, current thinking is we anchor trust to the root signer. rotate leaf only

jyao1: ok. then please document explicitly:
1) Intermediate cert check is ignored purposely.
2) Assumption: The leaf cert subject name contains enough information to unique identify the usage for this product / model.

sgrams:
@haitaohuang I reviewed the changes, PR looks good to me.
Can you please rebase and add code comments as requested by Jiewen? Thanks

src/crypto/src/lib.rs
haitaohuang: will do

haitaohuang:
comments added and rebased

jyao1 (approved):

PR #895, opened 2026-06-10
feat(migtd): route warn/error/fatal logs to COM1 serial port

### LINKED ISSUE #430: MigTD can use serial console to print log messages including warn/errors/fatal
Currently, debugging MigTD is quite hard. We had to put a debug TDVMCALL in order to see why MigTD is behaving in a certain way. This issue is making a suggestion to instead use VGA screen and/or serial console to print log messages including warn/error/fatal ones.

The debug messages can be part of the cfg and warn/error/fatals ones can always go to the VGA screen and/or serial port. 

MigTD can also use a shared buffer and then VMM can implement a TDVMCALL to get the PA of the string + length but this option requires changes in the VMM so instead a serial port maybe a better option. VGA screen buffer is small so maybe serial port is the best option if it feasible to implement in MigTD. The shared buffer is common whether we log it to serialport or VGA screen or over the TDVMCALL.

MichalTarnacki:
MigTD debugging has historically required ad-hoc TDVMCALL.IO writes inserted by hand, because failure paths often did not call any of the log::* macros and there was no guarantee that fatal exits reached the host console.

td-logger is already initialised with LevelFilter::Trace and, with the tdx feature, writes each byte to COM1 (0x3F8) via TDVMCALL.IO, which QEMU forwards through -serial mon:stdio. This change makes that serial path the canonical debugging channel:

Add release_max_level_info to the log dependency so release builds compile out debug!/trace! while keeping info!, warn! and error! on the serial line. No VMM changes, no shared buffer and no extra TDVMCALL are required.
Emit log::error! from panic_with_guest_crash_reg_report with the error code, message and caller file:line so every fatal exit is visible on the serial port even when the surrounding code forgot to log.
Fixes #430 

Reopen #838

### DIFF
--- a/src/migtd/src/driver/vmcall_raw.rs
+++ b/src/migtd/src/driver/vmcall_raw.rs
@@ -16,8 +16,6 @@ pub fn vmcall_raw_device_init() {
 
 #[track_caller]
 pub fn panic_with_guest_crash_reg_report(errorcode: u64, msg: &[u8]) -> ! {
-    #[cfg(not(feature = "vmcall-raw"))]
-    let _ = errorcode;
     let location = core::panic::Location::caller();
     let file = location.file();
     let line = location.line();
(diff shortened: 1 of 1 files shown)


sgrams:
@MichalTarnacki can you please update this commit so it resolves warning raised by clippy?

>variables can be used directly in the `format!` string

Thanks!

PR #902, opened 2026-06-18
fix(migtd): verify info_hash instead of servtd_hash

haitaohuang:
The old verify_servtd_hash computed SHA384(info_hash || type || attr) and compared to init_servtd_hash. Per GHCI spec, the correct check is to compare SHA384(masked_tdinfo) directly against init_servtd_info_hash, removing the unnecessary extra hash layer.

Also renames verify_servtd_hash to verify_servtd_info_hash for clarity.

Co-developed-by: Copilot <223556219+Copilot@users.noreply.github.com>

### DIFF
--- a/src/migtd/src/mig_policy.rs
+++ b/src/migtd/src/mig_policy.rs
@@ -87,8 +87,6 @@ mod v2 {
     const SERVTD_ATTR_IGNORE_RTMR2: u64 = 0x100_0000_0000;
     const SERVTD_ATTR_IGNORE_RTMR3: u64 = 0x200_0000_0000;
 
-    const SERVTD_TYPE_MIGTD: u16 = 0;
-
     lazy_static! {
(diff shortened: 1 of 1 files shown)


haitaohuang:
@sgrams @jyao1 Could you review this? We tested with 1.33 module

jyao1:
OK. It seems misunderstanding on SERVTD_HASH v.s. SERVTD_INFO_HASH.

```
SERVTD_HASH Calculation
Concatenate SERVTD_INFO_HASH, SERVTD_TYPE and SERVTD_ATTR of each slot in a temporary buffer:
 SERVTD_INFO_HASH in bytes 47:0
 SERVTD_TYPE in bytes 49:48
 SERVTD_ATTR in bytes 57:50
```

`INIT_INFO_HASH contains the value of the Service TD binding entry’s INFO_HASH at TDH.MR.FINALIZE time.`

I think we should use SERVTD_INFO_HASH to validate INIT_INFO_HASH.

jyao1 (approved):

PR #914, opened 2026-06-25
test(migtd): add regression test for ExchangeInformation wire layout

kali834x:
This originally fixed ExchangeInformation serializing four bytes of uninitialized trailing padding to the migration peer (no repr(C), implicit padding after max_ver copied out by as_bytes()). The same fix landed on main in 4f33b65 (repr(C) plus an explicit zeroed reserved field), so after rebasing only the regression test is left: it asserts the struct stays 40 bytes and that the reserved bytes serialize as zero, catching any future layout or field-order change that would reintroduce implicit padding.

### DIFF
--- a/src/migtd/src/migration/session.rs
+++ b/src/migtd/src/migration/session.rs
@@ -1399,6 +1399,26 @@ mod test {
         assert!(matches!(result, Ok(6)));
     }
 
+    // `ExchangeInformation` is written to the migration peer byte-for-byte via
+    // `as_bytes()`. The four bytes after `max_ver` were implicit trailing
+    // padding that the serializer copied to the peer uninitialized; they are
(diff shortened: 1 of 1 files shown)


kali834x:
any update?

sgrams:
Hi, can you please rebase your PR? Thank you

kali834x:
Done, rebased onto current main. No conflicts, the patch is unchanged.

kali834x:
Rebased again. Main picked up the same fix in 4f33b65 (repr(C) plus the zeroed reserved field), so this now only carries the regression test pinning the 40-byte layout and zeroed trailing bytes. Fine to close if you'd rather not take the test.

sgrams (approved):

sgrams:
Thank you for your contribution!

PR #921, opened 2026-07-02
fix(migtd): store digests as tagged event bodies

### LINKED ISSUE #922: Premigration fails with generated fmspc policy - TlsStream error
**Describe the bug**

Premigration fails with generated fmspc policy - TlsStream error 

Image

**To Reproduce**

1. cd tools/migtd-policy-generator
2. cargo build
3. cd ../../
4. ./target/debug/migtd-policy-generator -o config/policy_pre_production_fmspc_tool.json --pre-production
5. bash sh_script/preparation.sh
6. cargo image --policy config/policy_pre_production_fmspc_tool.json --root-ca config/Intel_SGX_Provisioning_Certification_RootCA_preproduction.cer
7. cd sh_script/test
8. sudo pytest -k "pre_migration" --servtd_hash $(cargo hash --image /root/MigTD/target/release/migtd.bin)

**Expected behavior**
Premigration should be successful.

Image

**Additional context**
It works with policy_pre_production_fmspc.json located in repository.

sgrams:
Resolves #922

The RA-TLS certificate embeds the full cc-measurement event log via the EXTNID_MIGTD_EVENT_LOG extension. In the v1 (non-policy_v2) path, the event log entries for the migration policy and SGX root CA carried the raw artifact bytes as their tagged event body, so growing the pre-production policy JSON from ~48 KiB to ~59 KiB pushed the reassembled handshake Certificate message past rustls' MAX_HANDSHAKE_SIZE cap (0xffff, 65535 B), surfacing as a TlsStream error during the RA-TLS handshake over virtio-vsock.

Store only a 48-byte SHA-384 digest of the artifact as the tagged event body in:

  * v1 get_policy_and_measure       (migration policy)
  * v1 get_ca_and_measure           (SGX root CA)
  * v2 get_policy_issuer_chain_and_measure (issuer chain)

RTMR values are unchanged because write_tagged_event_log still receives the full artifact via hash_data. verify_event_log only replays event header digests, and parse_events discards the tag body for these event IDs, so verifier compatibility is preserved. The RA-TLS certificate size is now decoupled from policy size and stays well under the rustls handshake cap regardless of future policy growth.

### DIFF
--- a/src/migtd/src/bin/migtd/main.rs
+++ b/src/migtd/src/bin/migtd/main.rs
@@ -240,7 +240,20 @@ fn get_policy_and_measure(event_log: &mut [u8]) {
         }
     };
 
-    let event_data = policy;
+    // Store only the policy digest as the tagged event body; RTMR is still
+    // extended with the full policy bytes above via `hash_data`. Keeps the
+    // RA-TLS cert (which embeds the event log) under rustls' 64 KiB
+    // MAX_HANDSHAKE_SIZE regardless of policy growth.
+    let event_data = match event_log::calculate_digest(policy) {
+        Ok(d) => d,
+        Err(e) => {
+            log::error!("Failed to digest migration policy: {:?}\n", e);
+            panic_with_guest_crash_reg_report(
+                MigrationResult::InitializationError as u64,
+                b"Failed to digest migration policy",
+            );
+        }
+    };
 
     // Measure and extend the migration policy to RTMR
     let _ = event_log::write_tagged_event_log(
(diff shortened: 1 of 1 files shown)


jyao1:
There are 2 consumers for the policy data.

1) Peer MigTD

Since we have described "MigTD Policy V2 Exchange"
_The MigTD source and destination need to exchange policy during attestation phase. Because the Policy V2 includes collateral information, it might be too big to be fit into the session establishment protocol. As such, the session establishment protocol (such as TLS or SPDM) may include the hash of the policy instead of including the whole policy data.
The policy data can be transported independently with the session establishment protocol by using following structure_.
I think it is OK to change to hash.

2) 3rd party Attestor

That will require the RIM includes a full policy data for verification.
If the attestor has a clear policy data/hash mapping, then it is OK to change as well.

PR #932, opened 2026-07-03
fix(deps): apply security fixes to td-shim-AzCVMEmu

MichalTarnacki:
|---|---|---|---|---|---|---|
| 1 | Low | tdcall_servtd_rebind_approve indexes target_td_uuid/token without length check | <entry> -> tdcall_servtd_rebind_approve | Use .get(..) and return Err on short slices. | weakness | The defect is real; the emulated tdcall_servtd_rebind_approve indexes target_td_uuid[0..3] and run copy_from_slice(&rebind_session_token[..32]) with no length check, so a slice shorter than 4 u64 or 32 bytes panics (CWE-248). But it is not an exploitable true positive. This is AzCVMEmu emulation code, outside the production TDX TCB — in the shipped product this call is a hardware TDCALL, and the production wrapper in tdx.rs:952-958 already validates both lengths (target_td_uuid.len() != 4 \|\| rebind_session_token.len() != REBIND_SESSION_TOKEN_SIZE) and returns Err before any indexing. |
| 2 | Medium | free_shared_pages: Box::from_raw with caller-supplied num  --  Layout mismatch + ov | <entry> -> free_shared_pages | Store original Layout alongside allocation; dealloc with that. Return Err on overflow. | weakness | Split the finding into its two claims. The "panic on huge num" via checked_mul(num).expect() is a false positive: num is never attacker-chosen. It is derived internally as dma_size / PAGE_SIZE, and dma_size = align_up(buf.len()) where buf.len() is hard-capped at MAX_VSOCK_PKT_DATA_LEN (64 KiB) by the enqueue guard, so num ≤ 16 and 0x1000 * 16 cannot overflow usize. The huge-num path is mathematically unreachable.<br><br>The "Layout mismatch → heap UB" half is a real defect but only a weakness, not the T1/MUST-FIX it is labeled. Reconstructing the dealloc Box from the caller's num instead of the recorded allocation size is a genuine footgun in an unsafe fn, but it only bites if some caller frees with a different count than it allocated. No in-tree caller does that — every path (the vsock Allocator, and virtio_pci storing dma_size and freeing with dma_size / PAGE_SIZE) is self-consistent. Crucially, the VMM does not control the free num; it controls bounded packet contents/lengths, while the page count is internal allocation bookkeeping. So the "directly reachable from untrusted VMM / oob_write" tag is overstated. This is also emulation-only code (td-payload-emu), outside the production TDX TCB. |
| 3 | Low | register_interrupt_callback truncates usize vector to u8 | <entry> -> register_interrupt_callback | Change parameter type to u8 or assert vector<256. | weakness | The vector parameter is always supplied by MigTD's own code (hardcoded interrupt vector constants passed to register_interrupt_callback), not by any external or VMM-controlled input. The truncation from usize to u8 is only dangerous if a caller passes a value ≥ 256, which no current caller does. The guard converts a hypothetical future misuse into a clean error return rather than silent index wraparound, but today the dangerous path is unreachable — making this defense-in-depth hardening, not remediation of an exploitable vulnerability. |

### DIFF
--- a/deps/td-shim-AzCVMEmu/td-payload-emu/src/arch/idt/mod.rs
+++ b/deps/td-shim-AzCVMEmu/td-payload-emu/src/arch/idt/mod.rs
@@ -5,6 +5,8 @@
 use interrupt_emu as intr;
 pub use intr::InterruptStack;
 
+const IDT_ENTRY_COUNT: usize = u8::MAX as usize + 1;
+
 #[derive(Copy, Clone)]
 pub struct InterruptCallback(fn(&mut InterruptStack));
 
@@ -18,6 +20,10 @@ impl InterruptCallback {
 }
 
 pub fn register_interrupt_callback(vector: usize, cb: InterruptCallback) -> Result<(), ()> {
+    // Validate vector fits in u8 for the IDT registry.
+    if vector >= IDT_ENTRY_COUNT {
+        return Err(());
+    }
     // Store raw fn into interrupt-emu registry
     intr::register(vector as u8, cb.0);
     Ok(())
--- a/deps/td-shim-AzCVMEmu/td-payload-emu/src/mm/shared/mod.rs
+++ b/deps/td-shim-AzCVMEmu/td-payload-emu/src/mm/shared/mod.rs
@@ -4,9 +4,17 @@
 // SPDX-License-Identifier: BSD-2-Clause-Patent
 
 use alloc::boxed::Box;
+use alloc::collections::BTreeMap;
 use alloc::vec::Vec;
+use spin::Mutex;
 
 const PAGE_SIZE: usize = 0x1000;
+
+/// Records the exact byte size allocated for each shared-page address so that
+/// `free_shared_pages` deallocates with the original `Layout` rather than one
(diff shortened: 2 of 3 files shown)


deps/td-shim-AzCVMEmu/td-payload-emu/src/arch/idt/mod.rs
```diff
@@ -18,6 +18,10 @@ impl InterruptCallback {
 }
 
 pub fn register_interrupt_callback(vector: usize, cb: InterruptCallback) -> Result<(), ()> {
+    // Validate vector fits in u8 (IDT has 256 entries)
+    if vector >= 256 {
```
sgrams: can you please replace with a define instead of hardcoding magic number?

sgrams (changes requested):

deps/td-shim-AzCVMEmu/td-payload-emu/src/arch/idt/mod.rs
MichalTarnacki: fixed

sgrams (approved):

PR #934, opened 2026-07-03
fix(migtd): wipe migration key material from TLS and SPDM buffers

MichalTarnacki:
|---|---|---|---|---|---|---|
| 1 | Medium | ExchangeInformation::as_bytes leaks 4B uninit padding to remote peer over TLS | exchange_msk -> exchange_info -> ExchangeInformation::as_bytes -> ratls_write | Add #[repr(C)] and zero-init via Default/zeroed(); or serialise field-by-field. | true_positive | Because #[derive(Default)] only initializes the named fields and size_of::<Self>() includes the alignment padding, from_raw_parts reads those bytes off the stack and ratls_client.write() sends them to the peer, which is a genuine CWE-908 uninitialized-memory disclosure. The severity is genuinely low, since the recipient is an already-attested, authenticated peer MigTD and the leaked bytes are version-adjacent padding rather than key material, but it is a real leak and not a false positive. |
| 2 | Medium | TLS plaintext buffers retain decrypted MSK after session  --  no zeroize on drop | exchange_msk -> SecureChannel::read -> TlsConnection::read -> ChunkVecBuffer::consume | impl Drop for ChunkVecBuffer/TlsBuffer that zeroizes inner Vec; or wrap in Zeroizing<Vec<u8>>. | true_positive | The decrypted MSK genuinely flows through two plaintext buffers that were never wiped: rustls decrypts records in place inside input: TlsBuffer (whose reset() only set used = 0 and discard() left stale plaintext in the tail, with no Drop), and received_app_data: ChunkVecBuffer stores payload.to_vec() copies that consume() dropped via pop_front()/drain() without zeroizing. So the 32-byte AES-256 key really did linger in freed heap and buffer tails, which is a real CWE-226, and it is inconsistent with the project's own intent since MigrationSessionKey already derives Zeroize/ZeroizeOnDrop.<br><br>MigTD's heap is TD-private encrypted memory the VMM cannot read, and virtio descriptors point at shared memory, so exploitation needs a separate in-TD memory-read primitive — making this defense-in-depth (effectively T4) rather than remotely reachable |
| 3 | Medium | app_context_data_buffer.zeroize() skipped on peer-triggered error path | spdm_responder_transfer_msk -> rsp_handle_message | Use a scopeguard/defer or Zeroizing<> wrapper so zeroize runs on all exit paths including `?`. | true_positive | In spdm_rsp.rs:150 the ? on rsp_handle_message(spdm_responder).await? genuinely returns before the app_context_data_buffer.zeroize() on the next line, so a peer that drives the responder into an Err after libspdm's signing callback has written the ephemeral RA-TLS private key into that buffer leaves the key material un-wiped, which is a real CWE-460 missing-cleanup. It is clearly a defect rather than intended behavior, because the codebase already applies the correct wrapper/inner pattern with unconditional zeroize for the requester in spdm_rebind.rs:24, and the same bug exists a second time in spdm_responder_rebind_new, so it is an inconsistency the authors simply missed on the responder side. One technical correction to the finding: at the encode() call the buffer holds PrivateKeyDer::default() (empty), not the P-384 key; the real key only enters the buffer later when libspdm's signing callback runs during message handling, so the leak window is during rsp_handle_message, but the mechanism and the fix still hold. |

### DIFF
--- a/src/migtd/src/migration/session.rs
+++ b/src/migtd/src/migration/session.rs
@@ -94,11 +94,25 @@ lazy_static! {
     pub static ref REQUESTS: Mutex<BTreeSet<u64>> = Mutex::new(BTreeSet::new());
 }
 
+#[repr(C)]
 #[derive(Default)]
 pub struct ExchangeInformation {
+    pub key: MigrationSessionKey,
     pub min_ver: u16,
     pub max_ver: u16,
-    pub key: MigrationSessionKey,
+    // Explicit zeroed padding field.
+    pub reserved: [u8; 4],
+}
+
+impl ExchangeInformation {
+    fn validate(&self) -> Result<()> {
+        if self.reserved != [0; 4] {
+            log::error!("ExchangeInformation::validate: reserved field must be zero\n");
+            return Err(MigrationResult::InvalidParameter);
+        }
+
+        Ok(())
+    }
 }
 
 #[cfg(not(feature = "spdm_attestation"))]
@@ -863,6 +877,7 @@ async fn migration_src_exchange_msk(
         log::error!(migration_request_id = info.mig_info.mig_request_id; "exchange_msk(): Incorrect ExchangeInformation size Size - Expected: {} Actual: {}\n", size_of::<ExchangeInformation>(), size);
         return Err(MigrationResult::NetworkError);
     }
+    remote_information.validate()?;
     shutdown_transport(ratls_client.transport_mut(), info.mig_info.mig_request_id).await?;
     Ok(())
 }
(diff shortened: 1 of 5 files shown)


src/migtd/src/migration/session.rs
```diff
@@ -94,11 +94,14 @@ lazy_static! {
...
 
+#[repr(C)]
 #[derive(Default)]
 pub struct ExchangeInformation {
+    pub key: MigrationSessionKey,
     pub min_ver: u16,
     pub max_ver: u16,
-    pub key: MigrationSessionKey,
+    // Explicit zeroed padding field.
+    pub reserved: [u8; 4],
```
sgrams, line 104: if this is explicitly meant to be zeroed, it also requires a check if reserved field is zeroed.

sgrams (changes requested):

src/migtd/src/migration/session.rs
MichalTarnacki: done

sgrams (approved):

PR #935, opened 2026-07-03
fix(migtd): harden virtqueue, vsock, SPDM and policy against untruste…

MichalTarnacki:
Commit ##1

|---|---|---|---|---|---|---|
| 1 | Medium | MigtdTransport::receive over-reads past frame boundary on stream transport | rsp_handle_message -> process_message -> SpdmDeviceIo::receive -> MigtdTransport::receive -> AsyncRead::read | First loop must read into &mut buffer[recvd..VMCALL_SPDM_MESSAGE_HEADER_SIZE]; second loop into &mut buffer[recvd..payload_size+12]; return payload_size+12 not recvd. | true_positive | The first loop reads into &mut buffer[recvd..] (the whole remaining buffer), so on a stream transport a coalesced read can return more than 12 bytes — recvd can already exceed payload_size+12. The second loop then never executes (or under-reads), and it returns recvd, handing the next frame's bytes to spdmlib as trailing garbage and desyncing the stream. Both reads must be capped: header to [recvd..12], payload to [recvd..payload+12], and return payload+12 |
| 2 | Medium | get_tcb_level_by_svn requires exact isvsvn match | <entry> -> TdIdentity::get_tcb_level_by_svn | Use highest level with isvsvn<=svn (floor match) per Intel TCB-Info semantics. | true_positive | The original exact-match (isvsvn == svn) misuses Intel TCB-Info/Enclave-Identity semantics, where tcbLevels are descending breakpoints and a reported svn must floor-match to the highest level with isvsvn <= svn; treating it as an exact lookup returns None for any svn between or above listed levels, which both rejects valid newer peers (availability) and lets an in-between svn skip the mandatory Revoked deny that it should have floored into (security). The floor-match fix is the correct, spec-conformant behavior and is backward-compatible for exact hits, so the finding is a genuine semantic defect. |
| 3 | Medium | json_set_signature re-parses+re-serializes via serde_json::Value before embeddin | <entry> -> json_set_signature | Sign raw input bytes; embed signature without re-serializing the data block. | weakness | The behavior is real, but it is fail-closed and only affects tooling: canonicalization is required so the same JSON always maps to the same bytes before signing/embedding. With the guard in json_sign_detached, non-canonical input is rejected early, and verification uses the exact embedded bytes. |
| 4 | Low | check-then-act across two separate spin::Mutex lock scopes | <entry> -> wait_for_request | Hold single lock across contains+insert; or use BTreeSet::insert return value. | weakness | There is one global EXECUTOR driven by a single poll_tasks() loop on the main thread, and the up-to-12 "concurrent" sessions are merely interleaved at .await points; since there is no .await between contains() and insert(), the whole window runs inside one task poll and no other session can interleave. REQUESTS is TD-private memory and request_id comes from the already-snapshotted private_mem, so the VMM cannot race or influence it. |
| 5 | Low | USED_PORT.lock().contains() then separate .lock().insert() in VsockStream::bind  --  latent split | VsockStream::bind | Hold single guard: `let mut g=USED_PORT.lock(); if g.contains(..){Err} else {g.insert(..);Ok}`. | weakness | The split-lock is real — bind and get_unused_port released the USED_PORT guard between contains() and insert(), so a concurrent binder could race into the gap and two streams could share a local port, colliding in BINDING_PKT_QUEUES. But in the shipped single-vCPU cooperative executor there is no preemption and no .await between the check and the insert, so there is no interleaving point and it is unreachable today; the finding itself rates it T5/internal/HARDENING. |
| 6 | Low | Rebind responder: op 0x03 (no variant check) -> op 0x05 (no variant check) -> write_msk(default) | spdm_responder_rebind_new -> migtd_vdm_msg_rsp_dispatcher_ex -> handle_exchange_mig_attest_info_req -> migtd_vdm_msg_rsp_dispatcher_ex -> handle_exchange_mig_info_req -> write_msk | Gate both migration-mode handlers on ResponderContextExInfo::MigrationInformation variant. | weakness | The op 0x05 handler handle_exchange_mig_info_req lacked the MigrationInformation variant guard that its three siblings have, and because vdm_message_transcript_before_finish is shared across modes, a legitimate rebind-attest could satisfy its only precondition — letting op 0x05 run in rebind mode against the default (zeroed) migration_info. However it stops short of a real key compromise: in rebind mode the buffer holds MigtdMigrationInformation::default(), so verify_servtd_attr(binding_handle=0, default_uuid) issues tdcall_servtd_rd against a non-bound handle and returns an attribute mismatch (or TDX error), failing before write_msk. So the chain is blocked by verify_servtd_attr, making it a weakness rather than a true positive — the added variant guard closes the mode-confusion gap early and restores symmetry with the other handlers, which is the correct hardening regardless. |
| 7 | Low | SPDM migration responder stores peer-supplied servtd_ext BEFORE attestation veri |  | Review the finding context and add appropriate input validation or guard checks. | weakness | In the production path the same peer-supplied servtd_ext bytes are verified inside rsp_verify_peer_attestation_v2; if that fails, op 0x03 returns Err, op 0x05 never runs, and write_approved_servtd_ext_hash is never reached, so no attacker-chosen hash ever reaches TDCS. The "write attacker-chosen APPROVED_SERVTD_EXT_HASH" primitive only materializes when attestation verification is compiled out (AzCVMEmu/test_mock_report/use-mock-quote), which are non-production test paths with no hostile peer. The original "store-before-verify" ordering was therefore a robustness/hardening defect rather than an exploitable flaw. |
| 8 | Info | AzCVMEmu copy_to_private_shadow returns alias to buf  --  no snapshot copy | SharedMemory::copy_to_private_shadow | Clone into a separate shadow Vec<u8> in the emu stub to preserve API contract parity with real td-payload, even if not strictly required for safety. | weakness | In AzCVMEmu builds buf is a process-private heap Vec with no concurrent host writer and the executor is single-threaded cooperative, so returning Some(&self.buf) instead of a copy grants no exploitable TOCTOU — the finding itself rates it Info/HARDENING/T5. The only real gap is fidelity: real td-payload returns a private snapshot decoupled from later shared-memory writes, while the emu stub returned a live alias, so a test relying on snapshot semantics could behave differently than production. That makes it a robustness weakness, not a vulnerability, and the trivial emu-only fix (copy into a separate shadow Vec) is justified for parity even though no security issue exists. |

commit ##2

|---|---|---|---|---|---|---|
| 3 | High | v2 evaluate_policy_forward/backward/common return Ok(()) when block is None | -> PolicyData::evaluate_policy_forward | Treat None as Err(PolicyIncomplete) or require at least one constraint per direction. | true_positive | The core claim is correct: evaluate_policy_forward/backward/common returned Ok(()) when the policy block was None, so an absent block silently skipped mandatory TCB and engine checks — a genuine fail-open that the design guide's acceptance procedure does not sanction (absence-permissive rules exist only for fmspc and the CRL numbers, never for TCB status or engine). However, the finding's RATLS_PEER_CERT / T1 "directly reachable from untrusted peer" framing is wrong: the evaluated policy is the local, signature-verified get_verified_policy(), not peer-supplied input, so it is not remotely exploitable and not Critical. |

### DIFF
--- a/deps/td-shim-AzCVMEmu/td-payload-emu/src/mm/shared/mod.rs
+++ b/deps/td-shim-AzCVMEmu/td-payload-emu/src/mm/shared/mod.rs
@@ -18,6 +18,7 @@ static SHARED_ALLOC_SIZES: Mutex<BTreeMap<usize, usize>> = Mutex::new(BTreeMap::
 
 pub struct SharedMemory {
     buf: Vec<u8>,
+    shadow: Vec<u8>,
 }
 
 impl SharedMemory {
@@ -29,6 +30,7 @@ impl SharedMemory {
         let size = pages.checked_mul(4096)?;
         Some(Self {
             buf: Vec::from_iter(core::iter::repeat(0u8).take(size)),
+            shadow: Vec::new(),
         })
     }
 
@@ -42,9 +44,11 @@ impl SharedMemory {
     }
 
     pub fn copy_to_private_shadow(&mut self) -> Option<&[u8]> {
-        // In emulation mode, just return the buffer directly since we're not dealing with
-        // actual shared/private memory conversion like in real TDX
-        Some(&self.buf)
+        // Copy into a separate shadow buffer so the returned slice is an immutable
+        // snapshot, matching the real td-payload shared->private memcpy semantics.
+        self.shadow.clear();
+        self.shadow.extend_from_slice(&self.buf);
+        Some(&self.shadow)
     }
 }
 
--- a/src/devices/vsock/src/stream.rs
+++ b/src/devices/vsock/src/stream.rs
@@ -109,11 +109,11 @@ impl VsockStream {
     }
 
     pub fn bind(&mut self, addr: &VsockAddr) -> Result {
-        if USED_PORT.lock().contains(&addr.port()) {
+        let mut used_port = USED_PORT.lock();
+        if !used_port.insert(addr.port()) {
             return Err(VsockError::AddressAlreadyUsed);
         }
-
-        USED_PORT.lock().insert(addr.port());
+        drop(used_port);
         self.addr.local.set_port(addr.port);
         Ok(())
     }
@@ -457,14 +457,16 @@ lazy_static! {
 }
 
 pub fn get_unused_port() -> Option<u32> {
-    let mut port = UNUSED_PORT_COUNTER.lock().checked_add(1)?;
+    let mut counter = UNUSED_PORT_COUNTER.lock();
+    let mut used_port = USED_PORT.lock();
+    let mut port = counter.checked_add(1)?;
 
-    while USED_PORT.lock().contains(&port) {
+    while used_port.contains(&port) {
         port = port.checked_add(1)?;
     }
 
-    USED_PORT.lock().insert(port);
-    *UNUSED_PORT_COUNTER.lock() = port;
+    used_port.insert(port);
+    *counter = port;
 
     Some(port)
 }
--- a/src/migtd/src/mig_policy.rs
+++ b/src/migtd/src/mig_policy.rs
@@ -286,16 +286,12 @@ mod v2 {
         let servtd_ext_src_obj =
             ServtdExt::read_from_bytes(servtd_ext_src).ok_or(PolicyError::InvalidParameter)?;
         let init_td_info = verify_init_tdinfo(init_tdinfo, &servtd_ext_src_obj)?;
-        let _engine_svn = policy
-            .servtd_tcb_mapping
-            .get_engine_svn_by_measurements(&Measurements::new_from_bytes(
-                &init_td_info.mrtd,
-                &init_td_info.rtmr0,
-                &init_td_info.rtmr1,
-                None,
-                None,
-            ))
-            .ok_or(PolicyError::SvnMismatch)?;
+
+        // migtdengine check: enginesvnsrc >= enginesvninit
+        policy.servtd_tcb_mapping.check_engine_not_older(
+            &engine_measurements_of(&tdx_report.td_info),
+            &engine_measurements_of(&init_td_info),
+        )?;
 
         // If backward policy exists, evaluate the migration src based on it.
         let relative_reference = get_local_tcb_evaluation_info()?;
@@ -541,6 +537,16 @@ mod v2 {
             u64::from_le_bytes(servtd_ext.init_attr),
             &servtd_ext.init_servtd_info_hash,
         )
+    }
+
+    /// `getengineinfo` measurement scope: the design guide classifies a MigTD
+    /// engine by `mrtd||rtmr[0-1]` only.
+    fn engine_measurements(mrtd: &[u8], rtmr0: &[u8], rtmr1: &[u8]) -> Measurements {
+        Measurements::new_from_bytes(mrtd, rtmr0, rtmr1, None, None)
+    }
+
+    fn engine_measurements_of(td_info: &TdInfo) -> Measurements {
+        engine_measurements(&td_info.mrtd, &td_info.rtmr0, &td_info.rtmr1)
     }
 
     fn get_rtmrs_from_tdinfo(
@@ -891,17 +897,18 @@ mod v2 {
                 ServtdExt::read_from_bytes(servtd_ext_src).ok_or(PolicyError::InvalidParameter)?;
             let init_td_info = verify_init_tdinfo(init_tdinfo, &servtd_ext_obj)?;
 
-            // Allowlist gate: init MigTD measurements must be in servtd_tcb_mapping
-            let _engine_svn = policy
+            let slice = |range: core::ops::Range<usize>| {
+                suppl_data.get(range).ok_or(PolicyError::InvalidParameter)
+            };
+            let peer = engine_measurements(
+                slice(Report::R_MIGTD_MRTD)?,
+                slice(Report::R_MIGTD_RTMR0)?,
+                slice(Report::R_MIGTD_RTMR1)?,
+            );
+            // migtdengine check: enginesvnsrc >= enginesvninit
+            policy
                 .servtd_tcb_mapping
-                .get_engine_svn_by_measurements(&Measurements::new_from_bytes(
-                    &init_td_info.mrtd,
-                    &init_td_info.rtmr0,
-                    &init_td_info.rtmr1,
-                    None,
-                    None,
-                ))
-                .ok_or(PolicyError::SvnMismatch)?;
+                .check_engine_not_older(&peer, &engine_measurements_of(&init_td_info))?;
         }
 
         Ok(suppl_data)
--- a/src/migtd/src/migration/session.rs
+++ b/src/migtd/src/migration/session.rs
@@ -599,11 +599,10 @@ pub async fn wait_for_request() -> Result<MigrationInformation> {
                     .ok_or(MigrationResult::InvalidParameter)?;
             let request_id = mig_info.mig_info.mig_request_id;
 
-            if REQUESTS.lock().contains(&request_id) {
-                Poll::Pending
-            } else {
-                REQUESTS.lock().insert(request_id);
+            if REQUESTS.lock().insert(request_id) {
                 Poll::Ready(Ok(mig_info))
+            } else {
+                Poll::Pending
             }
         } else if wfr.operation == 0 {
             Poll::Pending
--- a/src/migtd/src/spdm/mod.rs
+++ b/src/migtd/src/spdm/mod.rs
@@ -85,7 +85,7 @@ impl<T: AsyncRead + AsyncWrite + Unpin + Send> SpdmDeviceIo for MigtdTransport<T
         while recvd < VMCALL_SPDM_MESSAGE_HEADER_SIZE {
             let n = self
                 .transport
-                .read(&mut buffer[recvd..])
+                .read(&mut buffer[recvd..VMCALL_SPDM_MESSAGE_HEADER_SIZE])
                 .await
                 // ConnectionAborted maps from VmcallRawError::VmmCanceled.
                 // The SpdmDeviceIo trait cannot represent this error, so just
@@ -108,10 +108,11 @@ impl<T: AsyncRead + AsyncWrite + Unpin + Send> SpdmDeviceIo for MigtdTransport<T
             return Err(0_usize);
         }
 
-        while recvd < payload_size + VMCALL_SPDM_MESSAGE_HEADER_SIZE {
+        let total = payload_size + VMCALL_SPDM_MESSAGE_HEADER_SIZE;
+        while recvd < total {
             let n = self
                 .transport
-                .read(&mut buffer[recvd..])
+                .read(&mut buffer[recvd..total])
                 .await
                 .map_err(|e| {
                     if e.kind() == rust_std_stub::io::ErrorKind::ConnectionAborted {
@@ -122,7 +123,7 @@ impl<T: AsyncRead + AsyncWrite + Unpin + Send> SpdmDeviceIo for MigtdTransport<T
             recvd += n;
         }
 
-        Ok(recvd)
+        Ok(total)
     }
 
     async fn flush_all(&mut self) -> SpdmResult {
--- a/src/migtd/src/spdm/spdm_rsp.rs
+++ b/src/migtd/src/spdm/spdm_rsp.rs
@@ -149,15 +149,19 @@ pub fn spdm_responder<'a, T: AsyncRead + AsyncWrite + Unpin + Send + Sync + 'sta
     Ok((responder_context_ex, device_io_ref))
 }
 
-pub async fn spdm_responder_transfer_msk(
-    spdm_responder_ex: &mut ResponderContextEx<'_>,
-    mig_info: &MigtdMigrationInformation,
+pub async fn spdm_responder_transfer_msk<'a>(
+    spdm_responder_ex: &mut ResponderContextEx<'a>,
+    mig_info: &'a MigtdMigrationInformation,
     #[cfg(feature = "policy_v2")] peer_data: Vec<u8>,
 ) -> Result<(), SpdmStatus> {
     #[cfg(not(feature = "policy_v2"))]
     let peer_data = Vec::new();
 
     spdm_responder_ex.peer_data = peer_data;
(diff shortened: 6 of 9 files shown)


psidor99:
Row 6 is sighting-16787 (CWE-367), and per the audit it's Root cause group RC-069 — the same root cause as sighting-16701 (CWE-823), which my PR (#986) fixes. They're the same underlying bug: recycle_descriptors trusting descriptor fields read live from host-writable shared DMA.

This branch shadows the topology (desc_next/desc_write/chain_len), which does resolve 16787 and the free-list half of 16701. But recycle_descriptors here still reads addr/len live:

    let addr = desc.addr.read();
    let buf_len = desc.len.read();

so the CWE-823 arbitrary-read half of 16701 (forged addr → recv_pkt slice_from_raw_parts) and 16681 (CWE-825, forged addr → free_dma_memory frees the live ring) stay open, and the device-reported len isn't clamped.

Since 16787 and 16701 are one root cause (RC-069), it's cleaner to close them with a single complete shadow. My PR shadows all four fields (addr/len/flags/next) plus in_use/is_head, so it covers 16787 + 16701 + 16681 and subsumes this row 6. Proposal: drop the virtqueue.rs hunk here and rebase on my PR; keep the vsock/SPDM/policy/json-signer/emu fixes in this one.

MichalTarnacki:
> Row 6 is sighting-16787 (CWE-367), and per the audit it's Root cause group RC-069 — the same root cause as sighting-16701 (CWE-823), which my PR (#986) fixes. They're the same underlying bug: recycle_descriptors trusting descriptor fields read live from host-writable shared DMA.
> 
> This branch shadows the topology (desc_next/desc_write/chain_len), which does resolve 16787 and the free-list half of 16701. But recycle_descriptors here still reads addr/len live:
> 
> ```
> let addr = desc.addr.read();
> let buf_len = desc.len.read();
> ```
> 
> so the CWE-823 arbitrary-read half of 16701 (forged addr → recv_pkt slice_from_raw_parts) and 16681 (CWE-825, forged addr → free_dma_memory frees the live ring) stay open, and the device-reported len isn't clamped.
> 
> Since 16787 and 16701 are one root cause (RC-069), it's cleaner to close them with a single complete shadow. My PR shadows all four fields (addr/len/flags/next) plus in_use/is_head, so it covers 16787 + 16701 + 16681 and subsumes this row 6. Proposal: drop the virtqueue.rs hunk here and rebase on my PR; keep the vsock/SPDM/policy/json-signer/emu fixes in this one.

yes, lets drop the  virtqueue.rs change here and add everything whats missing in yours PR

sgrams:
The fix is right and the design guide backs it:

*migtd-new in rebinding:*
```
* (enginedatesrc, enginesvnsrc) = getengineinfo(migpolicymy.servtdidentity, tdreportsrc.mrtd||rtmr[0-1])
* (enginesvninit, enginedateinit) = getengineinfo(servtdidentitymy, tdinfoinit.mrtd||rtmr[0-1])
c) migtdengine check
  1. check enginesvnsrc: enginesvnsrc >= enginesvninit
  2. check enginedatesrc: enginedatesrc >= enginedateinit
```

*migtd-d in migration:* identical, modulo an `enginesvminit` typo in the guide.

Pre-fix, both sites resolved `init_td_info → _engine_svn` and dropped it; there was no `enginesvnsrc` operand at all, so clause (c) was simply unimplemented. Your operand choices are spec-exact: local mapping for both sides, peer measurements from the verified report, and `mrtd||rtmr[0-1]` with `rtmr2/rtmr3 = None` (correct, not lazy — `rtmr2` carries the migpolicy hash, so including it would make engine identity policy-dependent, which contradicts the guide's scope). The behaviour-change note is also sanctioned by the general-principles section: *"this rule is hidden in the policy, even if it is not explicitly written. migtd shall check the peer and reject."*

Three suggestions.

**1. Migration now derives the peer engine SVN twice, at two different scopes.** In `authenticate_migration_source_with_init_tdinfo`, `setup_evaluation_data` already computed a peer SVN via `get_engine_svn_by_report`, which fills `rtmr2: Some(..), rtmr3: Some(..)`:

```rust
// servtd_collateral.rs:192
pub fn get_engine_svn_by_report(&self, report: &Report) -> Option<u16> {
    let measurements = Measurements { /* ... */
        rtmr2: Some(bytes_to_hex_string(...)),
        rtmr3: Some(bytes_to_hex_string(...)),
    };
```

`compare_measurements` treats a pinned `rtmr2`/`rtmr3` in the mapping as mandatory but a `None` *query* as an automatic reject (`(Some(_), None) => return false`). So against a mapping entry that pins either register, `evaluation_data_src.migtd_isvsvn` is `None` while your new lookup succeeds — one flow rejects the peer via `UnqualifiedMigTdInfo` and the other admits it. Shipped configs pin neither, so it's latent, but `src/policy/test/policy_v2/tcb_mapping.json` pins `rtmr3`.

Note this can't be fixed by reusing `evaluation_data_src.migtd_isvsvn` — `setup_evaluation_data` is called with the *peer's* `verified_policy` (`mig_policy.rs:332`), so that field comes from the peer's `servtd_tcb_mapping`, not ours. Your local-mapping recompute is the correct operand; it's the older helper that's off-spec. Suggest narrowing it to the guide's scope so there's one rule:

```rust
// servtd_collateral.rs — getengineinfo scope is mrtd||rtmr[0-1] only.
pub fn get_engine_svn_by_report(&self, report: &Report) -> Option<u16> {
    self.get_engine_svn_by_measurements(&Measurements::new_from_bytes(
        report.get_migtd_info_property(&MigTdInfoProperty::MrTd).ok()?,
        report.get_migtd_info_property(&MigTdInfoProperty::Rtmr0).ok()?,
        report.get_migtd_info_property(&MigTdInfoProperty::Rtmr1).ok()?,
        None,
        None,
    ))
}
```

That's its only caller, and it makes both derivations agree by construction. Then `enforce_engine_against_init` can take the mapping as a parameter instead of re-fetching `get_verified_policy()` (both call sites already hold `policy`), which also drops the unreachable `InvalidParameter` branch:

```rust
fn enforce_engine_against_init(
    mapping: &TdTcbMapping,
    peer: &Measurements,
    init: &Measurements,
) -> Result<(), PolicyError> {
    let peer_svn = mapping.get_engine_svn_by_measurements(peer).ok_or(PolicyError::SvnMismatch)?;
    let init_svn = mapping.get_engine_svn_by_measurements(init).ok_or(PolicyError::SvnMismatch)?;
    if peer_svn < init_svn {
        return Err(PolicyError::SvnMismatch);
    }
    Ok(())
}
```

**2. The `enginedate` argument is an assumption, not a derivation.** Structurally the date *is* a function of the SVN (`get_tcb_level_by_svn` → `TcbLevel { tcb_date, .. }`), but `svn_src >= svn_init ⟹ date_src >= date_init` additionally needs `tcbDate` to be non-decreasing in `isvsvn`. That's Intel's Enclave-Identity convention; nothing in the code enforces or validates it. Fine to rely on — worth stating in the doc comment as a documented assumption rather than a consequence, so a future non-monotonic mapping doesn't quietly void the reasoning.

**3. No test for the case being fixed.** 54 insertions, no coverage, and the only CI job that reaches the changed rebinding path is *Rebind Prepare (Mock Report)* — the *Skip RA* variants build with `test_disable_ra_and_accept_all` and never call `authenticate_rebinding_old`. CI green here means "no regression," not "downgrade rejected." Now that the comparison is an isolated function, a direct unit test is cheap:

```rust
#[test]
fn test_engine_downgrade_rejected() {
    // mapping: measurements_a -> isvsvn 1, measurements_b -> isvsvn 2
    // peer=a, init=b  => Err(SvnMismatch)   <-- the bug this commit fixes
    // peer=b, init=a  => Ok(())
    // peer=a, init=a  => Ok(())
    // peer unknown to mapping => Err(SvnMismatch)  <-- the behaviour change
}
```

The last case is the one your commit message calls out as a behaviour change, so pinning it in a test is what makes that intentional rather than incidental.

**Out of scope, flagging so it isn't assumed closed:** the general-principles section adds a separate migration rule — *"in migration use case, the migtd-s and migtd-d shall have the same migtdengine (the migpolicy can be different)... migtd shall check the peer and reject."* That's `peer == me` equality, distinct from the `src >= init` inequality here, and it's still unimplemented.

MichalTarnacki:
> The fix is right and the design guide backs it:
> 
> _migtd-new in rebinding:_
> 
> ```
> * (enginedatesrc, enginesvnsrc) = getengineinfo(migpolicymy.servtdidentity, tdreportsrc.mrtd||rtmr[0-1])
> * (enginesvninit, enginedateinit) = getengineinfo(servtdidentitymy, tdinfoinit.mrtd||rtmr[0-1])
> c) migtdengine check
>   1. check enginesvnsrc: enginesvnsrc >= enginesvninit
>   2. check enginedatesrc: enginedatesrc >= enginedateinit
> ```
> 
> _migtd-d in migration:_ identical, modulo an `enginesvminit` typo in the guide.
> 
> Pre-fix, both sites resolved `init_td_info → _engine_svn` and dropped it; there was no `enginesvnsrc` operand at all, so clause (c) was simply unimplemented. Your operand choices are spec-exact: local mapping for both sides, peer measurements from the verified report, and `mrtd||rtmr[0-1]` with `rtmr2/rtmr3 = None` (correct, not lazy — `rtmr2` carries the migpolicy hash, so including it would make engine identity policy-dependent, which contradicts the guide's scope). The behaviour-change note is also sanctioned by the general-principles section: _"this rule is hidden in the policy, even if it is not explicitly written. migtd shall check the peer and reject."_
> 
> Three suggestions.
> 
> **1. Migration now derives the peer engine SVN twice, at two different scopes.** In `authenticate_migration_source_with_init_tdinfo`, `setup_evaluation_data` already computed a peer SVN via `get_engine_svn_by_report`, which fills `rtmr2: Some(..), rtmr3: Some(..)`:
> 
> ```rust
> // servtd_collateral.rs:192
> pub fn get_engine_svn_by_report(&self, report: &Report) -> Option<u16> {
>     let measurements = Measurements { /* ... */
>         rtmr2: Some(bytes_to_hex_string(...)),
>         rtmr3: Some(bytes_to_hex_string(...)),
>     };
> ```
> 
> `compare_measurements` treats a pinned `rtmr2`/`rtmr3` in the mapping as mandatory but a `None` _query_ as an automatic reject (`(Some(_), None) => return false`). So against a mapping entry that pins either register, `evaluation_data_src.migtd_isvsvn` is `None` while your new lookup succeeds — one flow rejects the peer via `UnqualifiedMigTdInfo` and the other admits it. Shipped configs pin neither, so it's latent, but `src/policy/test/policy_v2/tcb_mapping.json` pins `rtmr3`.
> 
> Note this can't be fixed by reusing `evaluation_data_src.migtd_isvsvn` — `setup_evaluation_data` is called with the _peer's_ `verified_policy` (`mig_policy.rs:332`), so that field comes from the peer's `servtd_tcb_mapping`, not ours. Your local-mapping recompute is the correct operand; it's the older helper that's off-spec. Suggest narrowing it to the guide's scope so there's one rule:
> 
> ```rust
> // servtd_collateral.rs — getengineinfo scope is mrtd||rtmr[0-1] only.
> pub fn get_engine_svn_by_report(&self, report: &Report) -> Option<u16> {
>     self.get_engine_svn_by_measurements(&Measurements::new_from_bytes(
>         report.get_migtd_info_property(&MigTdInfoProperty::MrTd).ok()?,
>         report.get_migtd_info_property(&MigTdInfoProperty::Rtmr0).ok()?,
>         report.get_migtd_info_property(&MigTdInfoProperty::Rtmr1).ok()?,
>         None,
>         None,
>     ))
> }
> ```
> 
> That's its only caller, and it makes both derivations agree by construction. Then `enforce_engine_against_init` can take the mapping as a parameter instead of re-fetching `get_verified_policy()` (both call sites already hold `policy`), which also drops the unreachable `InvalidParameter` branch:
> 
> ```rust
> fn enforce_engine_against_init(
>     mapping: &TdTcbMapping,
>     peer: &Measurements,
>     init: &Measurements,
> ) -> Result<(), PolicyError> {
>     let peer_svn = mapping.get_engine_svn_by_measurements(peer).ok_or(PolicyError::SvnMismatch)?;
>     let init_svn = mapping.get_engine_svn_by_measurements(init).ok_or(PolicyError::SvnMismatch)?;
>     if peer_svn < init_svn {
>         return Err(PolicyError::SvnMismatch);
>     }
>     Ok(())
> }
> ```
> 
> **2. The `enginedate` argument is an assumption, not a derivation.** Structurally the date _is_ a function of the SVN (`get_tcb_level_by_svn` → `TcbLevel { tcb_date, .. }`), but `svn_src >= svn_init ⟹ date_src >= date_init` additionally needs `tcbDate` to be non-decreasing in `isvsvn`. That's Intel's Enclave-Identity convention; nothing in the code enforces or validates it. Fine to rely on — worth stating in the doc comment as a documented assumption rather than a consequence, so a future non-monotonic mapping doesn't quietly void the reasoning.
> 
> **3. No test for the case being fixed.** 54 insertions, no coverage, and the only CI job that reaches the changed rebinding path is _Rebind Prepare (Mock Report)_ — the _Skip RA_ variants build with `test_disable_ra_and_accept_all` and never call `authenticate_rebinding_old`. CI green here means "no regression," not "downgrade rejected." Now that the comparison is an isolated function, a direct unit test is cheap:
> 
> ```rust
> #[test]
> fn test_engine_downgrade_rejected() {
>     // mapping: measurements_a -> isvsvn 1, measurements_b -> isvsvn 2
>     // peer=a, init=b  => Err(SvnMismatch)   <-- the bug this commit fixes
>     // peer=b, init=a  => Ok(())
>     // peer=a, init=a  => Ok(())
>     // peer unknown to mapping => Err(SvnMismatch)  <-- the behaviour change
> }
> ```
> 
> The last case is the one your commit message calls out as a behaviour change, so pinning it in a test is what makes that intentional rather than incidental.
> 
> **Out of scope, flagging so it isn't assumed closed:** the general-principles section adds a separate migration rule — _"in migration use case, the migtd-s and migtd-d shall have the same migtdengine (the migpolicy can be different)... migtd shall check the peer and reject."_ That's `peer == me` equality, distinct from the `src >= init` inequality here, and it's still unimplemented.

I updated review with 2. and 3., I am not sure about the first recomendation. It is does not affect the fix of the "v2 evaluate_policy_forward/backward/common return Ok(()) when block is None" bug so I would rather note that recommendation and address it later

sgrams (approved):

PR #936, opened 2026-07-03
fix(migtd): fence insecure attestation test features from release builds

MichalTarnacki:
|---|---|---|---|---|---|---|
| 1 | Medium | AzCVMEmu / use-mock-quote skip REPORTDATA↔pubkey channel binding  --  quote relay | verify_peer_cert -> verify_signature -> verify_public_key | For AzCVMEmu, implement the binding against the emulated REPORTDATA instead of skipping; for use-mock-quote, gate behind #[cfg(test)] or compile_error in release. | weakness | The code in server_client.rs:1107 really does early-return Ok(()) and skip the REPORTDATA↔SHA384(pubkey) channel binding under both AzCVMEmu and use-mock-quote, so the described bypass mechanism is real and the finding is not a false positive. But it is not a true positive either, because neither feature is in the default set and both are explicitly marked test-only in Cargo.toml:85; a normally shipped firmware build performs the full check, so there is no remotely reachable exploit in any released artifact — the attack requires someone to deliberately compile a dangerous test feature. |
| 2 | High | INV-5 violated: RA-bypass features are plain Cargo features, no release fence | exchange_msk -> ratls::client -> verify::verify_peer_cert | Add `#[cfg(all(feature="test_disable_ra_and_accept_all", not(debug_assertions)))] compile_error!(...)` for all three bypass features; or gate the mod behind #[cfg(test)]. | weakness | The finding is a weakness, a defense-in-depth hardening gap. The mechanism is real: under test_disable_ra_and_accept_all the entire verify module in server_client.rs:1176 returns Ok(()) unconditionally for verify_peer_cert and both rebinding verifiers, and the flag is an ordinary opt-in Cargo feature rather than a #[cfg(test)] gate, so accept-all RA-TLS genuinely exists in the build graph. The feature is not in default = [], it is explicitly marked "# Dangerous", and emits a runtime warning; exploitation requires a builder to deliberately pass the flag, which is the definition of a build-time defense-in-depth concern (the finding itself tags it T4, so the MUST-FIX bucket is overstated). |


sgrams (commented):
I think eventually we need both debug and release targets tested in the CI system. Let's merge this one and create an issue to track release target enablement.

sgrams (approved):

PR #943, opened 2026-07-07
fix(vsock): packet/credit/dump validation from security audit

MichalTarnacki:
|---|---|---|---|---|---|---|
| 1 | Medium | vsock Packet::check validates header len but not data_len vs buffer | <entry> -> check | In check(), require HEADER_LEN+data_len()<=buf.len(); return Err otherwise. | true_positive |  |
| 2 | Low | vsock credit counters rx_cnt/tx_cnt unchecked u32 add (wrap) | <entry> -> recv_packet_connected | Use wrapping_add intentionally per virtio-vsock spec, but move rx_cnt update AFTER data_len<=recv.len() check. | true_positive |  |
| 3 | Low | Slice index panic on VMM-controlled response length: Response::new() (vmcall.rs: |  | Review the finding context and add appropriate input validation or guard checks. | weakness | The reply.data()[0], reply.data()[1], and reply.data()[4..12] accesses can panic if the VMM returns a response with length == 24 (valid per Response::new() but leaving data() empty). The impact is exclusively a panic (DoS). The response buffer resides in shared memory fully controlled by the VMM, which already has unlimited DoS capability over MigTD through scheduling denial, VMCALL non-response, or simply returning VMCALL_STATUS_RESERVED indefinitely. |
| 4 | High | INV-1 violated: VMM shared-mem parsed in-place, not copied to private | handle_pre_mig -> vsock_transport_dequeue -> vmcall_service_migtd_receive -> Response::new -> Response::data | Copy response[..length] into private Vec<u8> immediately after Response::new validation; parse from private copy. Same fix for vmcall_raw process_buffer and session.rs:746. | true_positive | Both the send and receive poll_fn sanity checks read reply.data()[0], data()[1] and data()[4..12] with no prior length check, so a VMM setting the response length field to 24 makes data() empty and data()[0] panic the guest - a genuine host-triggered denial of service reachable directly from untrusted VMM input on every vmcall-vsock request. The broader "parse-in-place / re-read across polls" framing is weaker (the real payload is copied into a private Vec in recv_packet and bound to the request via mid), but the concrete out-of-bounds panic is a real, exploitable defect |

### DIFF
--- a/src/devices/vsock/src/transport/vmcall.rs
+++ b/src/devices/vsock/src/transport/vmcall.rs
@@ -26,6 +26,8 @@ const COMMAND_RECV: u8 = 4;
 const MAX_VSOCK_MTU: usize = 0x1000 * 16;
 const VMCALL_COMMON_HEADER_LEN: usize = 36;
 const VMCALL_STATUS_RESERVED: u32 = 0xffff_ffff;
+// Reply data layout: [0] version, [1] command, [2..4] reserved, [4..12] MigRequestId
+const VMCALL_REPLY_DATA_HEADER_LEN: usize = 12;
 const VMCALL_VECTOR: u8 = 0x52;
 pub(crate) const MAX_VSOCK_PKT_DATA_LEN: usize =
     MAX_VSOCK_MTU - VMCALL_COMMON_HEADER_LEN - HEADER_LEN;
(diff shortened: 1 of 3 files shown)


src/devices/vsock/src/transport/vmcall.rs
```diff
@@ -193,7 +193,8 @@ async fn vmcall_service_migtd_send(
         }
 
         // Do the sanity check
-        if reply.guid() != VMCALL_SERVICE_MIGTD_GUID.as_bytes()
+        if reply.data().len() < 12
```
sgrams: please replace magic number with a define

src/devices/vsock/src/transport/vmcall.rs
```diff
@@ -231,7 +232,8 @@ async fn vmcall_service_migtd_receive(
...
```
sgrams: please replace magic number with a define

sgrams (changes requested):

src/devices/vsock/src/transport/vmcall.rs
MichalTarnacki: done

src/devices/vsock/src/transport/vmcall.rs
MichalTarnacki: done

sgrams:
Can you please rebase this PR?

MichalTarnacki:
rebased

sgrams (approved):

PR #970, opened 2026-07-17
fix(migtd/migration): restrict GetTDReport to emulation builds

psidor99:
GetTDReport returns a MAC-valid TDREPORT over a VMM-chosen REPORTDATA, giving the host VMM a signing oracle over MigTD's identity (sighting-16709). Combined with the rebinding TDREPORT relay this enables unauthenticated rebinding approval on a same-platform NEW MigTD (sighting-16711).

No production migration or rebinding flow requires the VMM to invoke GetTDReport; all legitimate TDREPORT generation is MigTD-initiated during the RA-TLS handshake. Gate the request dispatch on the AzCVMEmu emulation feature and reject the operation as UnsupportedOperationError in production builds, removing the oracle at its source.

Assisted-by: GitHub Copilot:claude-opus-4-8

### DIFF
--- a/src/migtd/src/bin/migtd/main.rs
+++ b/src/migtd/src/bin/migtd/main.rs
@@ -541,6 +541,7 @@ fn handle_pre_mig() {
                             );
                             REQUESTS.lock().remove(&rebinding_info.mig_request_id);
                         }
+                        #[cfg(feature = "AzCVMEmu")]
                         WaitForRequestResponse::GetTdReport(wfr_info) => {
                             log::trace!(migration_request_id = wfr_info.mig_request_id; "Processing GetTdReport request\n");
                             let status = get_tdreport(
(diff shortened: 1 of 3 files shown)


sgrams:
Approach looks correct - GetTDReport is not part of any normative migration or rebinding flow in the Design Guide; all legitimate TDREPORT generation is MigTD-initiated during RA-TLS/SPDM attestation. Gating it behind AzCVMEmu at compile time is the right call.

One suggestion for defense-in-depth: consider also #[cfg(feature = "AzCVMEmu")]-gating the WaitForRequestResponse::GetTdReport enum variant in data.rs and the corresponding handler in main.rs. That way the type system guarantees the code path is unreachable in production — if someone accidentally adds a new dispatch path in the future, the compiler will refuse to build rather than silently allowing it.

psidor99:
Done. Check "Policy v2 Policy + TCB Mapping Leaf Rotation (Mock Report)" failed on "Setup build environment". I am not able to rerun this.

MichalTarnacki (approved):

sgrams (approved):

PR #986, opened 2026-07-27
fix(virtio/virtqueue): authenticate descriptor completions

psidor99:
The merged descriptor shadow reaps buffers from a private copy, but still
trusts the device-written used-ring id: pop_used never checked it was a slot
that was actually lent out, so a malicious VMM could complete a never-issued
slot, replay a slot to double-free it, complete a mid-chain (non-head) slot,
or over-report the written length.

Track outstanding descriptors with two driver-owned u32 bitmaps over the
existing shadow array (no heap allocation):

- in_use marks slots lent to the device; is_head marks each chain head.
- recycle_descriptors requires each slot to be in range and in_use, clearing
  both bits as it reaps, rejecting forged, stale or duplicate ids.
- pop_used requires the id to be in range, in_use and a genuine chain head
  before touching any descriptor, and clamps the device-reported len to the
  driver-recorded chain length.

A compile-time assert guards that MAX_QUEUE_SIZE stays within the 32-bit
bitmaps. No on-wire or memory layout change. Enforces that VMM input is
untrusted and checked before use (CWE-823).

Add unit tests for normal completion, out-of-range id, forged not-in-use id,
duplicate completion, mid-chain completion, corrupted shared descriptors,
and oversized device len.

Fixes: sighting-16701 (INV-3, CWE-823)
Fixes: sighting-16681 (CWE-825)
Assisted-by: GitHub Copilot (Claude Opus 4.8)

### DIFF
--- a/src/devices/virtio/src/virtqueue.rs
+++ b/src/devices/virtio/src/virtqueue.rs
@@ -13,6 +13,12 @@ use crate::{Result, VirtioError, VirtioTransport, PAGE_SIZE};
 
 const MAX_QUEUE_SIZE: usize = 32;
 
+// The `in_use` / `is_head` bitmaps are `u32`, so they can only track 32 slots.
+const _: () = assert!(
+    MAX_QUEUE_SIZE <= 32,
+    "in_use/is_head bitmaps are u32; widen them if MAX_QUEUE_SIZE grows"
+);
+
 #[derive(Debug)]
 pub struct VirtqueueBuf {
     pub addr: u64,
@@ -56,6 +62,10 @@ pub struct VirtQueue {
     /// addr/len/flags/next from this private shadow instead, which neutralizes
     /// host forgery of desc.addr/len/next.
     desc_shadow: [DescriptorShadow; MAX_QUEUE_SIZE],
+    /// Bitmap of outstanding descriptor slots.
+    in_use: u32,
+    /// Bitmap of slots that are the head of an outstanding chain.
+    is_head: u32,
 }
 
 impl VirtQueue {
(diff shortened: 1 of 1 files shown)


sgrams (approved):

MichalTarnacki:
I see your implementation is a bit different from the merged #989, it adds is_head /in_use bitmaps and unit tests; i think we need these changes; can you rebase this PR?

psidor99:
Done.

MichalTarnacki:
Also handles 

|---|---|---|---|---|---|---|
| 1 | Medium | recycle_descriptors chain-walk via shared desc.next/flags -> recycle never-submitted slots, leak in-flight | VirtioVsock::pop_used_rx -> VirtQueue::pop_used -> VirtQueue::recycle_descriptors | Store the chain (head,len,next-links) in private memory at add() time and replay from there during recycle; ignore shared desc.next/flags. | weakness | The mechanism is real: recycle_descriptors walks the chain by reading host-writable desc.next/desc.flags from shared DMA, and the chain head comes from the host-controlled used-ring id, so the VMM can redirect the walk to never-submitted slots, drift num_used/free_head, and leak descriptors until I/O stalls. But it is not an enforceable security violation in the TDX threat model: the index is always bounds-checked via get_mut, the loop is bounded by num_used so there is no OOB or infinite loop, and forged addr/len are validated downstream in recv_pkt against the TD-private dma_record. The only un-neutralized effect is bookkeeping drift leading to descriptor exhaustion, which is a denial-of-service — and DoS from the VMM, which owns the entire virtio-vsock transport anyway, is out of model. So it is a legitimate robustness weakness worth hardening. |

previously in #935

PR #988, opened 2026-07-27
fix(vsock/transport): free DMA records with matching allocator

psidor99:
The Drop implementation for VirtioVsock freed every entry in dma_record with the hardcoded free_shared_pages, while packet buffers are allocated through the injectable dma_allocator (VsockDmaPageAllocator) and released that way in free_dma_memory. On any platform whose allocator is not the shared-page allocator, this deallocator mismatch corrupts allocator state (CWE-590, finding 16762).

Free each dma_record entry with the allocator that produced it: the virtqueue ring pages, allocated directly via alloc_shared_pages so the ring is always host-visible independent of the injected allocator, are freed with free_shared_pages; packet buffers are freed via dma_allocator.free_pages, mirroring free_dma_memory. The ring base is tracked in a new queue_dma_pages field to distinguish it in Drop.

Assisted-by: GitHub Copilot:claude-opus-4.8 [ragitnow]

### DIFF
--- a/src/devices/vsock/src/transport/virtio_pci.rs
+++ b/src/devices/vsock/src/transport/virtio_pci.rs
@@ -64,6 +64,8 @@ pub struct VirtioVsock {
     /// DMA record table
     dma_record: BTreeMap<u64, DmaRecord>,
     rx_buf_num: usize,
+    // Ring pages base; allocated via shared-page allocator, freed in `Drop`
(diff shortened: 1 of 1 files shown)


MichalTarnacki (approved):

MichalTarnacki:
This issue is already resolved by #989, so closing this pr

PR #1004, opened 2026-08-12
Revert "fix(migtd/migration): restrict GetTDReport to emulation builds"

sgrams:
This reverts commit 7b98d8019365fd45a849def191d7d9782e830903.

### DIFF
--- a/src/migtd/src/bin/migtd/cvmemu.rs
+++ b/src/migtd/src/bin/migtd/cvmemu.rs
@@ -15,7 +15,7 @@ use migtd;
 use migtd::driver::vmcall_raw::panic_with_guest_crash_reg_report;
 use migtd::migration::event;
 use migtd::migration::logging::{create_logarea, enable_logarea, init_vmm_logger};
-use migtd::migration::session::{exchange_msk, report_status};
+use migtd::migration::session::{exchange_msk, get_tdreport_reportdata, report_status};
 use migtd::migration::MigrationResult;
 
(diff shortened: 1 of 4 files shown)


jyao1:
For https://github.com/intel/MigTD/commit/7b98d8019365fd45a849def191d7d9782e830903, I think we should NOT blindly disable the interface and block the VMM get a fresh TDReport for further evaluation.
I feel the fix could to change the final REPORTDATA_final = Math (NONCE_vmm_input), e.g. REPORTDATA_final = "VmmInputToMigTd" || SHA384("VmmInputToMigTdForHash" || NONCE_input), to make sure the VMM cannot input arbitrary REPORTDATA, while VMM can still input something to verify the freshness.
@haitaohuang , please let us know if the direction is OK.

haitaohuang:
Thanks @jyao1 that should work for us. Please ensure the final fix also address the original security issue triggered the disabling.

jyao1:
It seems there is no requirement to have a VMM input NONCE.
As such, the simple fix is to just use a fixed string as REPORTDATA = "HostRequestedMigTDReport".

@haitaohuang , please confirm if that is OK.

haitaohuang:
@jyao1 Yes, confirmed no requirement for VMM to input Nonce. Thanks

sgrams:
I'll prepare a follow up commit

sgrams:
@jyao1, @haitaohuang please see and review the follow-up fix commit

haitaohuang:
LGTM
Thanks

jyao1 (approved):

Issue #930, opened 2026-07-02
Proposal: Make the TD Identity artifact optional


mingweishih:
## Summary

Remove the **TD Identity** artifact (`isvsvn → (tcb_date, tcb_status)`)
entirely and drive the MigTD portion of migration policy from the ISV
**SVN alone**. MigTD's runtime identity lookup becomes a single step:
`SERVTD_INFO_HASH → isvsvn` (TCB Mapping). The SVN→date/status
correspondence moves out of the runtime trust surface into
**policy-authoring time**.

## Motivation

* **A MigTD TCB date is redundant with its SVN.** A date is only more
  expressive than an SVN when it is *globally comparable* across components —
  that is why platform TCB uses dates ("everything before 2024-01-01 is
  stale" holds across the TDX module, microcode, etc.). MigTD's release
  cadence is **independent** of those Intel components, so a MigTD `tcb_date`
  carries no cross-component freshness meaning; it degenerates to a
  per-release monotonic ordinal — which the SVN already is, only
  non-canonical. The date is therefore pure redundancy.
* **SVN already is the security ordinal.** Any "modern enough" judgment a
  date expresses can be expressed directly as an SVN bar
  (`migtdIdentity.isvsvn >= N`, absolute; or `>= self` / `>= init`,
  relative). The SVN→date correspondence is not lost: it moves to
  policy-authoring time — an operator who used to write
  `tcbDate >= 2026-04-03` now writes `isvsvn >= N`, where `N` is the SVN cut
  around that date, chosen once, offline.
* **Simplification.** Deletes a second signed CFV artifact, its signature /
  verification path, and an SVN→date table from the runtime trust surface —
  leaving the MigTD identity lookup as one hash→SVN step.

### Assumption

This proposal rests on the premise that **MigTD SVN is bumped only for
security-relevant changes** (the same monotonic, security-bumps-only
discipline as the TDX-Module SVN): a new SVN means a security-relevant change;
perf/bug-fix releases do not bump it. If that release discipline does not
hold, an SVN bar is a weaker signal than a date and the argument above does
not stand — so the discipline should be treated as a requirement of this
design, not an assumption to be revisited later.

## Security: no rollback regression

Removing TD Identity does **not** weaken rollback defense — the date/status it
carried was never what resisted rollback. The minimum SVN bar in the policy,
measured into RTMR2, prevents substituting a MigTD whose hash maps to a lower
SVN: an attacker cannot lower the bar without changing the local MigTD's own
[...]

Issue #929, opened 2026-07-02
Proposal: Use CoRIM as the MigTD endorsement format


mingweishih:
## Summary

As a follow-up for #908, this proposal suggests changing the endorsement
format to the **CoRIM** (Concise Reference Integrity Manifest, [draft-ietf-rats-corim-10](https://www.ietf.org/archive/id/draft-ietf-rats-corim-10.html)),
distributed as a standalone signed `COSE_Sign1` document that endorses
`SERVTD_INFO_HASH → isvsvn`.

The alternative — defining a new bespoke JSON schema — incurs the same
migration cost without the interoperability benefits.

## Motivation

The hash-based endorsement proposed by #908 invalidates the existing
JSON-based endorsement schema. Given the inevitable breaking change, this
proposal suggests using a new endorsement that adopts the CoRIM format for the
following benefits:

1. **Standard, externally-consumable endorsement.** This allows the
   endorsement to be easily adopted by third parties, and reusable in the
   remote attestation flow.
2. **Signature hygiene + trust binding.** COSE_Sign1 + RFC 9360 `x5chain`
   embeds the signer's certificate chain in the document itself, replacing
   hex-in-JSON signatures and the separate issuer-chain PEM. This pairs
   naturally with the root+subject signer anchor from #916.

-----

## Update

Adopting the feedback: rather than replacing the JSON endorsement with CoRIM,
the TCB Mapping will be offered in **two interchangeable formats**, selected at
**build time**:

* **JSON (default).** The existing signed-JSON collateral, reshaped to the
  #908 one-hash form. No new dependency; keeps the current signing pipeline.
* **CoRIM (feature-gated).** A standalone signed `COSE_Sign1` CoRIM, for
  deployments that want a standard, externally-consumable endorsement.

Both formats carry the **same** hash-based endorsement
(`SERVTD_INFO_HASH → isvsvn`) with identical semantics and the same trust
binding (RTMR1 signer anchor), so a build can pick either without changing the
trust model.

## JSON TCB Mapping format (default)

One-hash shape — endorses the hash only, with no per-register (MRTD/RTMR0..3)
fields:

```json
{
  "id": "BB9668CA-4EE8-4523-941A-B3B03BE46E03",
  "version": 1,
  "issueDate": "2025-01-01T00:00:00Z",
  "nextUpdate": "2026-01-01T00:00:00Z",
  "svnMappings": [
[...]

Issue #916, opened 2026-06-26
Proposing  new RTMR1 Signer-Anchor Measurement


haitaohuang:
This proposal changes what MigTD measures into RTMR1. Today RTMR1 extends over the raw bytes of the policy/identity issuer certificate chain. This makes RTMR1 far more sensitive than the runtime trust model:  validate_peer_cert_chain  only requires a matching root CA (DER) and leaf Subject — not the leaf key or exact chain bytes.

This mismatch causes three problems: (1) routine leaf-key rotation churns RTMR1, forcing a new build and endorsement for a runtime no-op; (2) regional leaf certificates fragment RTMR1 into one value per region despite an identical trust anchor; (3) CoRIM-based Azure (MAA) attestation embeds the chain via  x5chain , so hashing the chain into RTMR1 duplicates it as a second source of truth.

Proposed fix: extend RTMR1 over a stable signer anchor  A = H("MIGTD-RTMR1-ANCHOR-V1" || R || S) , where  R = H(DER(root))  and  S = H(DER(leaf subject)) . The CFV still ships the full chain (unchanged); only the hashed input changes.

Benefits: no rotation churn, region-independent measurement, no CoRIM duplication, and RTMR1 sensitivity that exactly matches the runtime trust model. Root-CA changes remain visible. Scope: RTMR1 only.

This is a follow-up on #908 and the full proposal is in the second commit of PR #910

jyao1:
Is that related to https://github.com/intel/MigTD/pull/890?

Question: It seems no way to validate if the authorized data is signed with the latest key in attestation flow. Is there any concern if the data is signed with old key?

Also, please describe explicitly the reason to exclude the intermediate certs, and be aware of the risk.

haitaohuang:
> Is that related to [#890](https://github.com/intel/MigTD/pull/890)?
> 

Yes this is on top of #890 and #908

> Question: It seems no way to validate if the authorized data is signed with the latest key in attestation flow. Is there any concern if the data is signed with old key?

Yes, that is an intentional trade-off: if an existing signed hash needs be revoked, we need remove it from svnMappings in TCBMapping. Old key signed any unknown hash will be rejected by authentic MigTD and attestation services with endorsed TCBMapping. Revoked hashes will be rejected by authentic MigTD and attestation services with updated endorsed TCBMapping

haitaohuang:
Updated proposal to include mitigations.

mingweishih:
The proposal looks good to me.

[...]

Issue #908, opened 2026-06-24
Proposing a new TCBMapping design with one hash endorsement


haitaohuang:
Policy v2 measures the entire signed policy blob (which contains the TCB mapping) into RTMR2. This creates a circular dependency:  svnMappings  can't include RTMR2 because RTMR2 depends on the  mapping's content. Today's workaround keys  svnMappings  on a register subset  [MRTD, RTMR0, RTMR1] , which (1) forces the source MigTD to accept full init TDINFO from the untrusted VMM per  request, and (2) prevents the attestation service from matching  init/cur_servtd_info_hash  against the mapping.

 Propose: measure RTMR2 as a single canonical-bytes extend of  policyData  with  servtdCollateral.servtdTcbMapping  redacted. The redaction is the only escape hatch; every other field stays bound by  construction. The mapping can then key on the full composite  tdinfo_hash  (=  SHA384(TDINFO) ), populated after measurement — no cycle.

 Benefits: MigTD resolves SVN locally (no VMM TDINFO), and the attestation service does a direct  tdinfo_hash → SVN  lookup with no out-of-band endorsements.  servtdIdentity  is retained for   tcbDate / tcbStatus  compatibility and measured for free. Future considerations: dropping policy signing and an RTMR1 signer anchor for key rotation.

See design doc at https://github.com/haitaohuang/MigTD/blob/tcbmap_proposal/doc/tcb_mapping_design_proposal.md

jyao1:
@haitaohuang , very good analysis!

In brief, the proposal is:

| MR       | Before                                  | After                                                    |
|---------|-------------------------------|------------------------------------------|
| MRTD  | TD-Shim + MigTDCore        | TD-Shim + MigTDCore  (unchanged) |
| RTMR0 | SEPARATOR                         | SEPARATOR (unchanged)                     |
| RTMR1 | **Policy issuer cert chain** | **TCBMapping issuer cert chain**        |
| RTMR2 | **Policy  **                          | **Policy   without TCBMapping**         |
| RTMR3 | N/A                                     | N/A (unchanged)                                  |

I agree that this solution (Using TDINFOHASH in the TCBMapping) can resolve the problem.

Some questions:
1) There is no authority for the policy, since policy issuer cert chain is not included in any RTMR. Then I do not see how signed policy is helping here. Do you still want signed policy, or do you want to propose for unsigned policy? 

[...]

PR #1021, opened 2026-09-01
support one-hash TCB mappings and CoRIM endorsements

haitaohuang:
This series implements the designs tracked by https://github.com/intel/MigTD/issues/908, https://github.com/intel/MigTD/issues/916, https://github.com/intel/MigTD/issues/929, and https://github.com/intel/MigTD/issues/930.


src/migtd/src/bin/migtd/main.rs
```diff
@@ -327,43 +330,52 @@ fn get_policy_and_measure(event_log: &mut [u8]) {
...
-                    b"Failed to compute policy signer anchor",
-                );
-            }
-        };
+    // RTMR1 is extended with the SHA384 signer anchor, not the entire PEM
+    // blob. The anchor is derived from the root CA fingerprint and the leaf
+    // certificate's dedicated signer EKU, with a domain-separation tag, so it
+    // is stable across PEM formatting changes and intermediate reissuance.
+    // When the 48-byte anchor is enrolled directly, `resolve_signer_anchor`
+    // returns it as-is, yielding the same RTMR1 value as the PEM-derived form.
```
jyao1, line 351: I think adding EKU is good. 
However, I am not convinced to replace subject check.

We should add all of them to ensure the security of key/cert update, including but not limited to
1) Root Anchor match (check Root-CA)
2) Purpose match (check key usage, EKU)
3) Identity continuity (check subject, Subject Alternative Name, Subject Distinguished Name, etc)
4) Anti-rollback (newer time/date)

Let's have a standalone PR to address this change.

haitaohuang: Requiring newer time/date complicates key rotation operation: we would still want migration happens during key rotation rollout period (note CRL will be implemented separately for key revoke).

haitaohuang:
this is replaced by #1032 and later series for corim support, etc. the anchor enforcement change will also be separate

PR #1032, opened 2026-09-03
feat: use complete TDINFO hash for TCB mappings

haitaohuang:
## Summary

Implement the one-hash TCB mapping design proposed in #908:

- use `SHA384(TDINFO)` over the complete unmasked TDINFO as the mapping key;
- break the RTMR2 circular dependency by measuring canonical `policyData` with only `servtdTcbMapping` redacted;
- verify the signed mapping with the RTMR1-bound policy issuer chain;
- authenticate migration and rebinding continuity through `SERVTD_EXT` rather than host-supplied Init_TDINFO;
- resolve source initial/current hashes through the authenticated source JSON mapping and reject lookup misses, SVN rollback, and nonzero `SERVTD_ATTR`;
- preserve cumulative mapping history in `migtd-hash`, including multiple release hashes assigned to the same SVN.

The legacy outer policy signature is ignored because policy integrity is established by RTMR2. The existing wire Init_TDINFO field remains for framing compatibility but is no longer used as authorization evidence.

Related to #908.

## Validation

- format, cargo check, clippy, cargo-deny, library build, and library tests;
- all 32 firmware build combinations and all six standalone tool builds;
- all 14 emulation workflow scenarios;
- all five workflows triggered on the fork branch: Format and Clippy, Library Crates, Fuzzing Test, main, and Integration (Emulation Mode).

## Future changes

Potential follow-up work, intentionally excluded from this focused series:

- replace raw policy-chain measurement with a stable RTMR1 trust anchor based on a root certificate and signer-purpose EKU;
- support direct trust-anchor enrollment and signer revocation through CRLs;
- support CoRIM TCB endorsements, including policies that use CoRIM without JSON servTD collateral;
- make servTD identity collateral optional where SVN-only evaluation is sufficient;
- add release tooling and coverage for independent policy, mapping, and identity signer rotation;
- strengthen two-phase measurement generation and hash-stability checks across release workflows.

PR #1035, opened 2026-09-04
feat(policy): add stable RTMR1 signer anchor and revocation

haitaohuang:
## Summary

Implements the RTMR1 signer-anchor proposal in #916 as a continuation of #1032:

- measure a stable signer anchor derived from the root certificate DER and leaf Subject DER instead of the raw policy issuer chain bytes;
- use the same anchor calculation in MigTD runtime measurement and `migtd-hash`;
- accept an authenticated servTD signer CRL and enforce it for local policy, identity, and mapping signer chains;
- apply the local CRL when validating peer signer chains during migration;
- expose a monotonic `servtdCrlNum` policy floor so peers cannot regress revocation state.

The full policy issuer chain remains enrolled in CFV and recorded in the event payload. This changes only the RTMR1 measurement input and signer-revocation enforcement. The existing JSON TCB mapping and TD Identity model from #1032 are retained.

This PR is stacked on #1032 and is intended as its focused continuation.

## Validation

- formatting, cargo check, clippy, cargo-deny, library build, and library tests;
- all 32 firmware image build combinations and all six standalone tool builds;
- all 14 emulation workflow scenarios;
- Azure policy-v2 IGVM package and hash generation.

## Follow-up work

See the future-work section in #1032 for the broader roadmap. Remaining related items identified while checking the TCB-mapping and signer-anchor work are:

- support CoRIM as an alternate TCB endorsement format, including direct trust-anchor enrollment and CoRIM-only policy builds (#929);
- authenticate and retain CoRIM `x5chain` material, apply the local servTD CRL to it, and consume authenticated peer CoRIM mappings during migration;
- reject CoRIM `nbf` and `exp` claims because MigTD does not have a trusted wall clock;
- make simplified TD Identity optional for SVN-only evaluation (#930);
- add release tooling and validation for independent policy, mapping, and identity signer rotation and measurement stability;
- strengthen peer policy error propagation and remove obsolete Init-TDINFO/local-TDINFO paths after the one-hash transition is complete.


jyao1:
Question: Is there any new test to validate the new revocation path?
 * `src/crypto/src/crl.rs` (`verify_crl_signature`, `is_serial_revoked`, `get_crl_issuer_der`)
 * `src/crypto/src/lib.rs` (`verify_signer_chain_not_revoked`/`verify_issuer_ca_constraints`)
 * `src/policy/src/v2/policy.rs` exercising `PolicyError::SignerRevoked` or `servtd_crl_num` evaluation
 * `src/migtd/src/mig_policy.rs` for the peer-chain-vs-local-CRL check.

jyao1:
On **Implicit constraint: the same CRL-issuing CA must be common to both chains it's checked against.** `servtd_crl` is a single field checked against both `issuer_chain` and `servtd_identity_issuer_chain`. `verify_signer_chain_not_revoked` independently looks for a CA cert matching the CRL's issuer inside whichever chain it's given, so this only works if both the policy-signer chain and the identity-signer chain actually share a common CA that issued the CRL. Reasonable for the expected deployment (single root/intermediate authority), but it's an assumption the code enforces implicitly (fails closed with "CRL issuer does not match any CA in the signer chain" if violated) rather than one stated anywhere near the `servtd_crl` field or in the design doc's schema section.
A one-line doc/comment note would save a future reader from re-deriving this.

jyao1:
Comment: **`RawPolicyData::verify`'s local-CRL peer check silently no-ops if the local policy has no CRL.** By design the CRL is optional, so this is expected, but it means revocation enforcement for peers is entirely dependent on whether the *verifying* side happens to carry a CRL — a peer's own compromised/revoked chain is otherwise accepted.
Already implied by "optional field" in the design doc; flagging only so it's an explicit, acknowledged operational requirement (ship a CRL if you want revocation enforced) rather than something a deployer discovers later.

src/policy/src/v2/measurement.rs
```diff
@@ -0,0 +1,401 @@
...
+    let root_hash = digest_sha384(root_der).map_err(|_| PolicyError::HashCalculation)?;
+    let subject_hash = digest_sha384(leaf_subject_der).map_err(|_| PolicyError::HashCalculation)?;
+
+    let mut input = Vec::with_capacity(
+        SIGNER_ANCHOR_DOMAIN_TAG.len() + root_hash.len() + subject_hash.len() + 2,
+    );
+    input.extend_from_slice(SIGNER_ANCHOR_DOMAIN_TAG);
+    input.push(SIGNER_ANCHOR_SEPARATOR);
+    input.extend_from_slice(&root_hash);
+    input.push(SIGNER_ANCHOR_SEPARATOR);
```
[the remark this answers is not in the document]
jyao1: I am fine to add separator.

haitaohuang:
> Question: Is there any new test to validate the new revocation path?
> 
> * `src/crypto/src/crl.rs` (`verify_crl_signature`, `is_serial_revoked`, `get_crl_issuer_der`)
> * `src/crypto/src/lib.rs` (`verify_signer_chain_not_revoked`/`verify_issuer_ca_constraints`)
> * `src/policy/src/v2/policy.rs` exercising `PolicyError::SignerRevoked` or `servtd_crl_num` evaluation
> * `src/migtd/src/mig_policy.rs` for the peer-chain-vs-local-CRL check.
Added now

haitaohuang:
> On **Implicit constraint: the same CRL-issuing CA must be common to both chains it's checked against.** `servtd_crl` is a single field checked against both `issuer_chain` and `servtd_identity_issuer_chain`. `verify_signer_chain_not_revoked` independently looks for a CA cert matching the CRL's issuer inside whichever chain it's given, so this only works if both the policy-signer chain and the identity-signer chain actually share a common CA that issued the CRL. Reasonable for the expected deployment (single root/intermediate authority), but it's an assumption the code enforces implicitly (fails closed with "CRL issuer does not match any CA in the signer chain" if violated) rather than one stated anywhere near the `servtd_crl` field or in the design doc's schema section. A one-line doc/comment note would save a future reader from re-deriving this.

Added documentation

haitaohuang:
> Comment: **`RawPolicyData::verify`'s local-CRL peer check silently no-ops if the local policy has no CRL.** By design the CRL is optional, so this is expected, but it means revocation enforcement for peers is entirely dependent on whether the _verifying_ side happens to carry a CRL — a peer's own compromised/revoked chain is otherwise accepted. Already implied by "optional field" in the design doc; flagging only so it's an explicit, acknowledged operational requirement (ship a CRL if you want revocation enforced) rather than something a deployer discovers later.

Added comments and fail close if CRL is missing.

src/crypto/src/crl.rs
```diff
@@ -120,9 +121,140 @@ pub fn get_crl_number(crl: &[u8]) -> Result<u32, Error> {
...
+pub fn is_serial_revoked(crl: &[u8], serial: &[u8]) -> Result<bool, Error> {
+    let der = crl_pem_to_der(crl)?;
+    let crl = Crl::from_der(&der).map_err(|_| Error::ParseCertificate)?;
+
+    Ok(crl
+        .tbs_cert_list
+        .revoked_certificates
+        .iter()
+        .flatten()
+        .any(|entry| entry.user_certificate.as_bytes() == serial))
```
ksayid, line 174: just wondering: should we reject delta CRLs here unless we also have the corresponding base CRL?
since a delta CRL only contains changes relative to the base, checking it directly with `is_serial_revoked()` could miss certificates that were revoked in the base CRL?

src/policy/src/v2/policy.rs
```diff
@@ -605,6 +615,13 @@ impl ServtdPolicy {
             }
         }
 
+        if let Some(property) = &self.servtd_crl_num {
+            let servtd_crl_num = value.servtd_crl_num.ok_or(PolicyError::CrlEvaluation)?;
+            if !property.evaluate_integer(servtd_crl_num, relative_reference.servtd_crl_num)? {
```
ksayid, line 620: just wondering: is servtdCrlNum guaranteed to always refer to the same CRL issuer/scope on both sides? I think CRL numbers are only comparable within the same stream. is that enforced somewhere?

src/crypto/src/lib.rs
```diff
@@ -136,6 +154,42 @@ pub fn verify_cert_chain_and_signature(
...
+            issuer_key = Some(extract_public_key_from_cert(cert)?);
+            break;
+        }
+    }
+    let issuer_key = issuer_key.ok_or_else(|| {
+        Error::CertChainVerification("CRL issuer does not match any CA in the signer chain".into())
+    })?;
+    crl::verify_crl_signature(crl_pem, &issuer_key)?;
+
+    for cert in &chain {
```
ksayid, line 182: just wondering: should the revocation lookup be scoped to certificates issued by the CRL issuer?

since serial numbers are only unique per issuer, could checking every cert in the chain against the CRL’s serials potentially reject an unrelated cert that happens to have the same serial? for example, if an intermediate CA has serial `2` from the root, and that intermediate’s CRL revokes one of its own leaf certs with serial `2`, could we incorrectly treat the intermediate itself as revoked?

src/crypto/src/lib.rs
```diff
@@ -136,6 +154,42 @@ pub fn verify_cert_chain_and_signature(
...
+
+    let crl_issuer_der = crl::get_crl_issuer_der(crl_pem)?;
+    let mut issuer_key = None;
+    for cert in &chain {
+        let subject_der = cert
+            .tbs_certificate
+            .subject
+            .to_der()
+            .map_err(|_| Error::ParseCertificate)?;
+        if subject_der == crl_issuer_der && is_ca_certificate(cert)? {
```
ksayid, line 172: just wondering: should we also check KeyUsage.cRLSign before accepting this cert as the CRL issuer? It looks like we currently require BasicConstraints.cA=true, but not that the certificate is authorized to sign CRLs

ksayid (approved):

src/crypto/src/crl.rs
haitaohuang: I don't think we will load delta CRLs because it's one time loading at migtd startup

src/policy/src/v2/policy.rs
haitaohuang: It must be issued by the same entity. This is expected to be enforced by MigTD owner / the policy signer

src/crypto/src/lib.rs
haitaohuang: yes this design expect MigTDs only trust peers of the same signer as measured by RTMR1. CRLs also signed by the same signer.

